PR #1, opened 2017-05-19
TRAC-9948 Ignore permission denied when attempting to find multipart uploads.

gnicol:
https://trac.cyberduck.io/ticket/9948

This approach is basically a copy/paste from S3Session's list method.

We still attempt to find any existing multipart uploads so we can
resume them. However, if we get permission denied we'll now just
log/ignore the error and carry on with a fresh upload.

This allows uploads to buckets where the user lacks the permission
s3:ListBucketMultipartUploads
This situation is expected for any multi-tenant bucket.


gnicol:
Thanks very much for accepting this PR; particularly for getting it in so promptly!

Apologies I didn't notice the duplicate Trac issues you had to link/close. I did search for existing issues but somehow missed the two you found :(

dkocher:
Thanks for your contribution. We have amended this fix with d19a22f.

PR #2, opened 2017-05-28
feature/9711-support-manta-object-store

tjcelaya:
PR is nearly ready, just fixing a few final test cases. Will call out `TODO`s that require attention in later comments.


dkocher:
@tjcelaya Could you prune the unused imports in your IDE to remove the code quality warnings.

manta/src/main/java/ch/cyberduck/core/manta/MantaBufferWriteFeature.java
```diff
@@ -0,0 +1,67 @@
...
+import ch.cyberduck.core.io.BufferInputStream;
+import ch.cyberduck.core.io.FileBufferSegmentingOutputStream;
+import ch.cyberduck.core.transfer.TransferStatus;
+
+import org.apache.commons.io.IOUtils;
+
+import java.io.IOException;
+
+/**
+ * TODO: not sure how what triggers usage of this class
```
dkocher: The implementation of this interface is not required, when the protocol does not have a multipart upload requirement for large files. This interface is only used in Mountain Duck, where we require a write feature with no prior knowledge of the final content length to be transferred.

manta/src/main/java/ch/cyberduck/core/manta/MantaExceptionMappingService.java
```diff
@@ -0,0 +1,115 @@
...
+            return mapHttpException(me, ObjectUtils.firstNonNull(me.getContent(), ""));
+        }
+
+        if(failure instanceof MantaIOException) {
+            return new DefaultIOExceptionMappingService().map((IOException) ExceptionUtils.getRootCause(failure));
+        }
+        return new InteroperabilityException(failure.getMessage(), failure);
+    }
+
+    public BackgroundException mapHttpException(final MantaClientHttpResponseException httpFailure, final String path) {
```
dkocher: You can possibly resuse `HttpResponseExceptionMappingService` with `return new HttpResponseExceptionMappingService().map(new HttpResponseException(httpFailure.getStatusCode(), buffer.toString()))`.

manta/src/main/java/ch/cyberduck/core/manta/MantaFindFeature.java
```diff
@@ -0,0 +1,79 @@
...
+
+    public MantaFindFeature(final MantaSession session) {
+        this.session = session;
+        this.cache = PathCache.empty();
+    }
+
+    @Override
+    public boolean find(final Path file) throws BackgroundException {
+
+        // TODO: the following block is duplicated in several other Find features?
```
dkocher: You can possibly omit this and just rely on the `AttributesFinder` implementation.

manta/src/main/java/ch/cyberduck/core/manta/MantaFindFeature.java
```diff
@@ -0,0 +1,79 @@
...
+        if(cache.isHidden(file)) {
+            // Previously not found
+            return false;
+        }
+        // END duplicate code cache usage
+
+        try {
+            final PathAttributes found = new MantaAttributesFinderFeature(session).withCache(cache).find(file);
+            return true; // successfully found attributes for file
+        } catch (BackgroundException e) {
```
dkocher: Replace with `NotfoundException`.

manta/src/main/java/ch/cyberduck/core/manta/MantaObjectAttributeAdapter.java
```diff
@@ -0,0 +1,107 @@
...
+    private static final Logger log = Logger.getLogger(MantaAttributesFinderFeature.class);
+
+    private final MantaSession session;
+
+    public MantaObjectAttributeAdapter(MantaSession session) {
+        this.session = session;
+    }
+
+    public PathAttributes from(final MantaObject mantaObject) {
+        // TODO: directoryId for cryptomator?
```
dkocher: Must not be set here.

manta/src/main/java/ch/cyberduck/core/manta/MantaTouchFeature.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    @Override
+    public boolean isSupported(final Path workdir) {
+        return session.pathMapper.isUserWritable(workdir);
+    }
+
+    @Override
+    public Touch withWriter(final Write writer) {
+        // TODO: same as withCache, would it help to use the writer?
```
dkocher: Not required as the implementation does not make use of the `Write` feature.

manta/src/main/java/ch/cyberduck/core/manta/MantaWriteFeature.java
```diff
@@ -0,0 +1,97 @@
...
+
+    @Override
+    public boolean random() {
+        return false;
+    }
+
+    // TODO: what does this checksum do? can we combine it with our MD5 checksum?
+    @Override
+    public ChecksumCompute checksum() {
+        // TODO: verify this is actually used
```
dkocher: The checksum will be set in the `TransferStatus#checksum` for use in the `Write` feature implementation if you want to set it in a HTTP header for example to allow the server to verify integrity.

core/src/main/java/ch/cyberduck/core/Protocol.java
```diff
@@ -99,7 +99,19 @@ public boolean validate(final Credentials credentials, final LoginOptions option
         },
         irods,
         b2,
-        file;
+        file,
+        manta {
+            @Override
+            public boolean validate(final Credentials credentials, final LoginOptions options) {
+                if(credentials.isAnonymousLogin()) {
```
dkocher: This check can be omitted.

manta/pom.xml
```diff
@@ -0,0 +1,57 @@
...
+            <artifactId>test</artifactId>
+            <type>pom</type>
+            <scope>test</scope>
+            <version>${project.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>com.joyent.manta</groupId>
+            <artifactId>java-manta-client</artifactId>
+            <version>LATEST</version>
```
dkocher: Set to latest stable version.

manta/src/main/java/ch/cyberduck/core/manta/MantaDirectoryFeature.java
```diff
@@ -0,0 +1,66 @@
...
+        return directory;
+    }
+
+    @Override
+    public boolean isSupported(final Path workdir) {
+        return session.pathMapper.isUserWritable(workdir);
+    }
+
+    /**
+     * TODO: what does this do?
```
dkocher: Empty implementation is just fine because this feature does not make use of `Write` interface.

manta/src/main/java/ch/cyberduck/core/manta/MantaBufferWriteFeature.java
tjcelaya: Is it possible to invoke this feature from the cyberduck cli presently? Manta does support a form of [server-side multipart upload](https://github.com/joyent/java-manta#server-side-multipart-upload) which _should_ make it possible to support that use-case.

dkocher:
Implementing is really only required when the service would *only* support multipart uploads for certain file sizes to make it work with Mountain Duck.

If you want to have support for multipart uploads in Cyberduck – to get better performance for large single file uploads – you can implement the `Upload` feature. See `SwiftLargeObjectUploadFeature`, `S3MultipartUploadService` and `B2LargeUploadService`.

dkocher:
Please merge with master to make the Travis CI build green.

manta/src/main/java/ch/cyberduck/core/manta/MantaSession.java
```diff
@@ -0,0 +1,192 @@
...
+                        .setHttpsProtocols(DefaultsConfigContext.DEFAULT_HTTPS_PROTOCOLS)
+                        .setDisableNativeSignatures(true)
+                        .setNoAuth(false)
+                        .setMantaURL("https://" + h.getHostname())
+                        .setMantaUser(host.getCredentials().getUsername()));
+    }
+
+    @Override
+    protected MantaClient connect(final HostKeyCallback key) throws BackgroundException {
+        return null;
```
dkocher: Should return the configured client.

manta/src/main/java/ch/cyberduck/core/manta/MantaSession.java
```diff
@@ -0,0 +1,192 @@
...
+    public MantaSession(final Host h) {
+        this(h, new DisabledX509TrustManager(), new DefaultX509KeyManager());
+    }
+
+    public MantaSession(final Host h, final X509TrustManager trust, final X509KeyManager key) {
+        super(h, new ThreadLocalHostnameDelegatingTrustManager(trust, h.getHostname()), key);
+
+        pathMapper = new MantaPathMapper(this);
+        exceptionMapper = new MantaExceptionMappingService(this);
+        config = new ChainedConfigContext(
```
dkocher: Please extend from `HttpSession` to make `MantaClient` use the `HttpConnectionPoolBuilder` configured with custom trust and key managers.

manta/src/main/java/ch/cyberduck/core/manta/MantaSession.java
tjcelaya: The `MantaClient` object isn't instantiated until the `login` call. We can't create one until we have the user's account name and key.

dkocher: Can't you use a `SettableConfigContext` you keep a reference to and update credentials later after instantiation in login?

tjcelaya: The `MantaClient` constructor expects to be passed a usable `ConfigContext` and will throw an exception for the missing parameter values:

```
Errors when loading Manta SDK configuration:
Manta key id must be specified
Either the Manta key path must be set or the private key content must be set. Both can't be null.
```

I've been meaning to propose a restructuring of the way configs are prepared since it's not obvious that configuration values passed to the `MantaClient` instance cannot be changed after initialization. May I ask about the purpose of returning the client from `connect(HostKeyCallback)`?

dkocher: The API expects the client instance to be instantiated and configured after `connect(HostKeyCallback)`. If the connection requires authentication it can set credentials in `login(LoginCallback…)`. A reference to the client instance is held in the super class `Session`.

manta/pom.xml
```diff
@@ -0,0 +1,57 @@
...
+  ~ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  ~ GNU General Public License for more details.
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <artifactId>parent</artifactId>
+        <groupId>ch.cyberduck</groupId>
+        <version>6.0.1-SNAPSHOT</version>
```
dkocher: Fix the parent POM reference to `6.1.1-SNAPSHOT` to make the build pass.

osx/pom.xml
```diff
@@ -330,5 +330,10 @@
             <artifactId>cryptomator</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>ch.cyberduck</groupId>
+            <artifactId>manta</artifactId>
+            <version>6.0.1-SNAPSHOT</version>
```
dkocher: Fix the dependency version.

dkocher:
Can you fix the `com.joyent.manta:java-manta-client` dependency.

manta/src/main/java/ch/cyberduck/core/manta/MantaMoveFeature.java
```diff
@@ -39,12 +40,8 @@ public MantaMoveFeature(MantaSession session) {
     }
 
     @Override
-    public void move(final Path file, final Path renamed, final boolean exists, final Delete.Callback callback) throws BackgroundException {
+    public void move(final Path file, final Path renamed, final TransferStatus transferStatus, final Delete.Callback callback) throws BackgroundException {
         try {
-            if(exists) {
```
dkocher: You can query `transferStatus.isExists()` if required.

dkocher:
I would love to pull this in – could you resolve the remaining issues with the build?

dkocher:
I would love to pull this in – could you resolve the remaining issues with the build?

`Tests in error: 
  MantaSessionTest.testFeatures:44 Â» IllegalArgument Username can't be empty
  MantaSessionTest.testSessionRejectsBadUsernames:73->assertUsernameFailsLogin:56 Â» IllegalArgument
`

tjcelaya:
@dkocher Apologies for the slowest PR ever, this is ready to be merged now.

PR #3, opened 2017-08-28
Set file modified date to src_last_modified_millis if found, follow up fixes to CD-2552

timdream:
This follow-ups b97bfd62 and checks src_last_modified_millis when listing and download files.

I have no experience in Java so I am not sure how to add any tests :(. Please add tests for me if you willing and find it necessary.


timdream:
BTW I've verified manually this little change fixes the file timestamp on the download and the listing case, as described in the support case, at the same time.

PR #5, opened 2017-11-01
Fixes for feature/TRAC-9711

tjcelaya:
This PR is the result of the latest round of acceptance testing and should make the Manta implementation release-ready. Issues addressed include:

General Improvements:
- Allow modifying the host field.
- Reorder `DefaultsConfigContext` usage, later configs take precedence and this was causing odd errors.
- Friendly error messages for login failures instead of `HTTP request failed: [1:mantaSdkVersion= blah]` and in case users attempt to list the manta root (i.e. the "Cannot list Buckets" error was not being displayed).
- Fix MD5 checksum conversion so it stops seeming like every single download fails to checksum.
- Public key authentication actually seems to interact with the KeyChain correctly now. Previously we were only setting the password into the `ConfigContext` if we prompted for the passphrase leading to a situation where users could only log in once.
- Stop providing the key as key content now that [joyent/java-manta #294](https://github.com/joyent/java-manta/issues/294) is resolved.
- Implemented the `Search` feature in a way that makes it useful since the default implementation would take ages to scan through the `~~/reports` and `~~/jobs` folders. I'd like some feedback on how (or if) it would be possible to deliver partial results from the user-writable directories and then continue searching the system directories.

Testing-related fixes:
- Integration tests continue to use the hard-coded RSA key and default host by default, but allow for overriding with `manta.key_path` and `manta.url` (so I can point the tests at a real Manta without interfering with the usual suite)
- un-`@Ignore`d `MantaPublicKeyAuthenticationTest` and switched to JUnit assumptions so they'll be skipped automatically if the relevant system properties are not specified
- Drop `MantaSessionTests#assertUsernameFailsLogin`

The only thing left is the release of `com.joyent.manta:java-manta-client-unshaded:3.2.0` which will occur tomorrow.


dkocher:
Thanks for the cleanup! Could you add an integration test for the newly added search feature?

tjcelaya:
Integration test was added. The Codacy build seems to be failing because the branch isn't being analyzed at all. Travis is failing because of a missing commit.

dkocher:
Please just annotate the test with `@Category(IntegrationTest.class)`. We do not run integration tests on feature branches.

tjcelaya:
Apologies, I totally forgot to add the `@Category(IntegrationTest.class)` annotation

PR #9, opened 2018-01-24
Add OCI Object Storage Profiles

itemir:
This PR adds connection profiles for Oracle OCI Object Storage


profiles/OCI Object Storage us-ashburn-1.cyberduckprofile
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>Protocol</key>
+	<string>s3</string>
+	<key>Vendor</key>
+	<string>oracle-us-phoenix-1</string>
```
dkocher: Vendor strings should be unique among all profiles.

profiles/OCI Object Storage eu-frankfurt-1.cyberduckprofile
```diff
@@ -0,0 +1,36 @@
...
+	</string>
+	<key>Region</key>
+	<string>eu-frankfurt-1</string>
+	<key>Regions</key>
+	<array>
+		<string>us-phoenix-1</string>
```
dkocher: As we are connecting to a non-global region endpoint it is probably not advisable to allow creating buckets for different regions – I make the assumption that these region specific requests for new buckets with a region different than the connected endpoint will fail?

dkocher:
Would it be possible that you embed an icon that is based on our [template](https://github.com/iterate-ch/cyberduck/blob/master/profiles/assets/Template.psd)?

profiles/OCI Object Storage us-ashburn-1.cyberduckprofile
itemir: This is a remnant from my previous testing. Do we need Regions key at all? When it is omitted, does Cyberduck pick the region from Region key while creating top level folders (i.e. buckets)?

profiles/OCI Object Storage eu-frankfurt-1.cyberduckprofile
itemir: This is a remnant from my previous testing. Do we need Regions key at all? When it is omitted, does Cyberduck pick the region from Region key while creating top level folders (i.e. buckets)?

itemir:
> Would it be possible that you embed an icon that is based on our template?

I am afraid it is beyond my image editing capabilities...

itemir:
> Would it be possible that you embed an icon that is based on our template?

I managed to do it but didn't look good when resized. So removed it altogether.

itemir:
Also removed the Regions key, looks like it is not required. 

Let me know how how the final version looks.

itemir:
@dkocher I added an icon based on the Cyberduck template and added the regions back. Please have a look and let me know if this looks good.

PR #14, opened 2018-10-29
Update Donate.strings ES

JorgeCoke:
Fix typo


dkocher:
Thanks for your contribution! We are managing translations through [Transifex](https://www.transifex.com/cyberduck/cyberduck/content/).

PR #173, opened 2019-10-14
Update dropbox@2x.png

majouji:
Update the Dropbox disk icon to more closely match Dropbox’s new branding.

## New Icon

dropbox@2x


majouji:
This can be used for Mountain Duck as well, but it looks like that project is not open source.

PR #358, opened 2020-07-15
Use STANDARD for default google storage class

itspage:
This should fix https://trac.cyberduck.io/ticket/11062

`multi_regional` is no longer recommended for use.
It is equivalent to standard but only works with Multi-Regional buckets.

See https://cloud.google.com/storage/docs/storage-classes#legacy


dkocher:
@dependabot merge

PR #415, opened 2020-10-29
Add SUFeedURL to Info.plist on macOS

NanoSector:
This fixes TRAC ticket https://trac.cyberduck.io/ticket/11219 .

The default url of https://version.cyberduck.io/local/changelog.rss built with this patch does not work - I assume this string will be left empty on release build as https://version.cyberduck.io//changelog.rss does work.


dkocher:
Previously deemed unnecessary in 607795f3.

PR #462, opened 2021-01-21
add more the one external editor on windows 

janbildsoehansen:
i don't know if i'm writing this the right place but would be nice to be able to add more then one editor on windows


dkocher:
This is the wrong place. Please write to support@cyberduck.io or open a new ticket for enhancement requests at https://trac.cyberduck.io/newticket

PR #528, opened 2021-05-31
Add possibility to override InstallFolder variable

antoine-aumjaud:
I would like to install Cyberduck by command line *without admin rights*.

When I try to launch this:
```
Cyberduck-Installer-7.9.1.34974.exe  /install /quiet /norestart /log log.txt InstallFolder=%USERPROFILE%/bin/cyberduck
```

I have in the log.txt file : 
```
 Initializing string variable 'InstallFolder' to value '[ProgramFiles64Folder]Cyberduck'
 [...]
 Ignoring attempt to set non-overridable variable: 'InstallFolder'.
```

I have added this flag `bal:Overridable="yes"` on Wix config files for Windows & CLI

### DIFF
--- a/cli/src/main/wix/Bootstrapper/Cyberduck CLI Bundle.wxs
+++ b/cli/src/main/wix/Bootstrapper/Cyberduck CLI Bundle.wxs
@@ -7,7 +7,7 @@
       <bal:WixStandardBootstrapperApplication LicenseUrl=""      
         LogoFile="$(var.SetupDir)banner.png" LogoSideFile="$(var.SetupDir)welcome.bmp" />
     </BootstrapperApplicationRef>
-    <Variable Name="InstallFolder" Type="string" Value="[ProgramFiles64Folder]Cyberduck CLI" />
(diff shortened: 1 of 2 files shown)


AliveDevil:
Hi Antoine,

thanks for your contribution, though unfortunately setting this property to overridable won't make this installer non-admin.
Following things would need to be satisfied to make the installer non-admin:
`windows\src\main\wix\Bundle` has to be modified, to remove `InstallScope="perMachine"` (this will render the Bootstrapper invalid, as it requires an InstallScope-parameter).
There are a bunch of hard coded registry values to HKLM and HKCR for File Associations, protocol handlers.

The bootstrapper automatically installs [Apple Bonjour](https://developer.apple.com/bonjour/) - no chance of getting that to install per-user. For a per-user installation this would have to be excluded.

It all boils down to this: [WiX - Does Burn support dual-purpose msi packages](https://stackoverflow.com/a/30626469).

For a general question though: You don't mind whether this is a MSI or MSIX-package, right?

We are discussing how to proceed here.

antoine-aumjaud:
Thanks for your quick answer.
MSI or MSIX-package is not a matter, I just want a package that can be installed without admin right, it could be just a zip :)

PR #531, opened 2021-06-03
Ignore broken directories in Cryptomator vaults

tbodt:
This is the same as the BrokenDirectoryFilter in the main Cryptomator
codebase.
https://github.com/cryptomator/cryptofs/blob/develop/src/main/java/org/cryptomator/cryptofs/dir/BrokenDirectoryFilter.java

### DIFF
--- a/core/src/main/java/ch/cyberduck/core/vault/DecryptingListProgressListener.java
+++ b/core/src/main/java/ch/cyberduck/core/vault/DecryptingListProgressListener.java
@@ -23,6 +23,7 @@ import ch.cyberduck.core.PathAttributes;
 import ch.cyberduck.core.Session;
 import ch.cyberduck.core.exception.BackgroundException;
 import ch.cyberduck.core.exception.ConnectionCanceledException;
+import ch.cyberduck.core.features.Find;
 import ch.cyberduck.core.features.Vault;
 
 import org.apache.log4j.Logger;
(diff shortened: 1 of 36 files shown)


tbodt:
https://drewdevault.com/2018/10/05/Dont-sign-a-CLA.html

core/src/main/java/ch/cyberduck/core/vault/DecryptingListProgressListener.java
```diff
@@ -54,6 +57,9 @@ public void visit(final AttributedList<Path> list, final int index, final Path f
                 }
                 list.set(index, vault.decrypt(session, f).withAttributes(new PathAttributes(f.attributes()).withVersions(versions)));
             }
+            if (find != null && list.get(index).isDirectory() && !find.find(vault.encrypt(session, list.get(index)))) {
```
dkocher, line 60: This will cause performance issues with large directory listings.

tbodt: Is there a better way to do this? Any directory that fails this check does not actually exist, and problems will appear later when you try to e.g. delete it.

dkocher: As this is the exception it would probably be a better approach to just ignore failures for missing files on delete silently.

tbodt: This is really a problem with listings and can still appear if you don't try to delete anything. If the directory shows up in a listing even though it doesn't exist, almost anything that interacts with it will fail, including listing - not just deleting.

dkocher:
My take on this is in https://github.com/iterate-ch/cyberduck/compare/bugfix/TRAC-11696?expand=1 which would ignore failures where the target directory referenced in metadata is missing.

PR #554, opened 2021-07-03
Support AWS SSO via CLI login

yuzawa-san:
This patch attempts to add support to read SSO credentials via the CLI cache. That is, the user may sign in using `aws sso login` and then use Cyberduck. I noticed that this application does not currently use the AWS SDK for Java v2 which support SSO. Additionally, the current implementation and v1 SDK does not look at the [config file](https://docs.aws.amazon.com/sdkref/latest/guide/creds-config-files.html) just the credentials file. For this reason and compatibility purposes, I altered the logic to coalesce the credentials and config profiles in a manner generally consistent with the AWS documentation. The latest version of the AWS CLI creates a config file and credentials file. Specifically in the SSO case, there is a config entry for the profile, but no entry in the credentials file. Upon `aws sso login`, the CLI writes the credentials to a cache. This implementation reads from that cache. I considered upgraded to AWS SDK v2, but it looked to difficult. If we upgraded, I believe the SDK would handle most of this (outside of the MFA case).

Testing: Use the cyberduck profile from https://trac.cyberduck.io/wiki/help/en/howto/s3#Connectingusingcredentialsin.awscredentials to specify the profile name. 

I believe the documentation may need to be updated to mention the SSO use case.


dkocher:
Thanks for your contribution! It may take some time on my side to review this on vacation. I would certainly prefer if we could do SSO without the dependency on the cached data from the `aws` tool but I agree that upgrading to _AWS SDK for Java v2_ is probably non trivial.

s3/src/main/java/ch/cyberduck/core/sts/STSCredentialsConfigurator.java
```diff
@@ -73,27 +88,42 @@ public STSCredentialsConfigurator(final X509TrustManager trust, final X509KeyMan
...
+                if(credentialsFileProperties != null){ 
+                    configFileProperties.putAll(credentialsFileProperties);
+                }
+                allProfileProperties.put(profileName, configFileProperties);
+            }
         }
-        catch(AccessDeniedException | IllegalArgumentException e) {
-            log.warn(String.format("Failure reading %s", file), e);
+        catch(AccessDeniedException | IllegalArgumentException | IOException e) {
+            log.warn(String.format("Failure reading %s and ", configFile, credentialsFile), e);
```
dkocher, line 120: Is there a `%s` missing in the string formatting?

dkocher (commented):
Sorry it has taken such a long time to add my review. I agree we can merge this and reevaluate a autonomous implementation independent from AWS CLI when we would migrate to AWS SDK for Java 2.0. We would switch only when we decide to ditch our usage of [JetS3t](https://github.com/iterate-ch/jets3t).

dkocher:
If you could rebase this to `master` and resolve the conflict we can merge this.

PR #12636, opened 2021-12-08
Bugfix/gh 12625

### LINKED ISSUE #12625: Bad Request. not a valid rfc 3339 formatted date
**Describe the bug**
Uploading with a timestamp set in transfer status leads to interoperability error.

Part of request with `content_modified_at`
```
--HHazpWpHIskW8FNx9gI-eUonGjuHaC4dw
Content-Disposition: form-data; name="attributes"
Content-Type: application/octet-stream
Content-Transfer-Encoding: binary

{"name":"cryptomator_256.png","parent":{"id":"150624862371"},"content_modified_at":"2021-12-06T11:13:51.042Z"}
```

Error response
```
{"type":"error","status":400,"code":"bad_request","help_url":"http:\/\/developers.box.com\/docs\/#errors","message":"not a valid rfc 3339 formatted date","request_id":"ktiqjogx4bskx0c8"}
```

bgkarthik27:


box/src/main/java/ch/cyberduck/core/box/BoxWriteFeature.java
```diff
@@ -69,32 +79,30 @@ public BoxWriteFeature(final BoxSession session, final BoxFileidProvider fileid)
...
-                    new JSON().getContext(null).writeValue(content, new FilescontentAttributes()
-                            .name(file.getName())
-                            .parent(new FilescontentAttributesParent().id(fileid.getFileId(file.getParent(), new DisabledListProgressListener())))
-                            .contentModifiedAt(status.getTimestamp() != null ? new DateTime(status.getTimestamp()) : null)
-                    );
+                    final FilescontentAttributes filescontentAttributes = new FilescontentAttributes()
+                        .name(file.getName())
+                        .parent(new FilescontentAttributesParent().id(fileid.getFileId(file.getParent(), new DisabledListProgressListener())))
+                        .contentModifiedAt(status.getTimestamp() != null ? new DateTime(status.getTimestamp()) : null);
+                    final byte[] content = getObjectMapper().writer().writeValueAsBytes(filescontentAttributes);
```
dkocher, line 99: We use `new JSON().getContext(null)` everywhere else to obtain the mapper. Why not use this and add the custom date time serializer module there?

dkocher (changes requested):

box/src/main/java/ch/cyberduck/core/box/BoxWriteFeature.java
bgkarthik27: If I use that, then it includes the registered Joda Module. I do not find any option to unregister that registered module.

dkocher (approved):

PR #12710, opened 2021-12-23
Add support reading `IdentityAgent` from OpenSSH configuration.

### LINKED ISSUE #10454: Interoperability with gpg-agent (YubiKey)
> **03b55c6** created the issue

Cyberduck is great, but I cannot use it anymore, since I switched to using a YubiKey to store my SSH private key.

The reason is simply **the requirement to select a private key file** when opening a connection.
_Note_: I can connect using SSH on the command-line, using ForkLift, SourceTree, … just fine. They all just use the identity agent I have configured…

Since this blocks me from using Cyberduck, I consider this a defect…

dkocher:
Resolves #10454.

### DIFF
--- a/ssh/src/main/java/ch/cyberduck/core/sftp/SFTPSession.java
+++ b/ssh/src/main/java/ch/cyberduck/core/sftp/SFTPSession.java
@@ -43,6 +43,7 @@ import ch.cyberduck.core.sftp.openssh.OpenSSHAgentAuthenticator;
 import ch.cyberduck.core.sftp.openssh.OpenSSHCredentialsConfigurator;
 import ch.cyberduck.core.sftp.openssh.OpenSSHHostnameConfigurator;
 import ch.cyberduck.core.sftp.openssh.OpenSSHIdentitiesOnlyConfigurator;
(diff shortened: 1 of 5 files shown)


AliveDevil (approved):

ylangisc (approved):

sindastra:
Hi there, are these changes already live? I'm asking because I downloaded Cyberduck just now (version 8.2.3), and this does not work for me.

lnhrdt:
Hey @dkocher I'm really excited for this feature. I've been wishing this type of OpenSSH configuration was supported for years, so thank you!

I saw in the release notes this was included with 8.2.2 but I'm using 8.2.3 and unable to figure out how to make it work. Is there anything special we need to do in the connection configuration, or otherwise? Or are there any limitations to what configurations are supported?

florisvdg:
Same for me, haven't been able to get this to work. Launching the app with `SSH_AUTH_SOCK` does work (already prior to this PR), but anything I try with `IdentityAgent` does not.

@dkocher Do you have an example config snippet that works for you?

PR #12785, opened 2022-01-13
Switch to BasicContextSelector due to class loading issues with IKVM …

ylangisc:
…on Windows.


dkocher (approved):

AliveDevil:
Do we want to have this fixed in https://github.com/iterate-ch/ikvm?

ylangisc:
> Do we want to have this fixed in https://github.com/iterate-ch/ikvm?

Wouldn't put time into this. It's probably more an issue how Log4j2 lookups contexts.

PR #12798, opened 2022-01-19
Allow glob pattern in --download argument for CLI

### LINKED ISSUE #12797: Unable to download files using glob pattern

dkocher:


ylangisc (approved):

core/src/main/java/ch/cyberduck/core/transfer/DownloadTransfer.java
```diff
@@ -199,6 +199,10 @@ public TransferAction action(final Session<?> source, final Session<?> destinati
                 final Local local = download.local;
                 if(local.exists()) {
                     if(local.isDirectory()) {
+                        if(download.remote.isDirectory()) {
+                            // Do not prompt when merging into existing directory
+                            continue;
```
AliveDevil, line 204: Why? There is `-e (overwrite|skip)` which does the exact same thing.

PR #12818, opened 2022-01-24
Switch to rollover strategy using 5 compressed archive files.

ylangisc:


core/src/main/csharp/ch/cyberduck/core/preferences/SettingsDictionaryPreferences.cs
```diff
@@ -147,17 +147,19 @@ protected override void configureLogging(String level)
             base.configureLogging(level);
             var fileName = Path.Combine(LogDirectoryFinderFactory.get().find().getAbsolute(),
```
dkocher, line 148: Extract var for `LogDirectoryFinderFactory`

dkocher (changes requested):

dkocher (approved):

PR #12872, opened 2022-02-11
Add parameter to choose between revealing and selecting file or show …

dkocher:
…browser rooted at folder instead.


ylangisc (approved):

ylangisc:
Not sure yet, if in case of a file it should be opened or selected only with `select=false`.

PR #12887, opened 2022-02-15
Fix copying objects when connected to single bucket

dkocher:


AliveDevil:
Not working.
grafik

s3/src/test/java/ch/cyberduck/core/s3/S3LocationFeatureTest.java
```diff
@@ -140,9 +140,14 @@ public void testEmptyThirdPartyProvider() {
...
         final Profile profile = new ProfilePlistReader(factory).read(
-            new Local("../profiles/Wasabi (us-central-1).cyberduckprofile"));
+                new Local("../profiles/Wasabi (us-central-1).cyberduckprofile"));
         final S3Session session = new S3Session(new Host(profile, profile.getDefaultHostname()));
         assertFalse(new S3LocationFeature(session).getLocations().isEmpty());
         assertTrue(new S3LocationFeature(session).getLocations().contains(new Location.Name("us-central-1")));
     }
+
+    @Test
+    public void testVirutalhost() {
```
AliveDevil, line 150:
```suggestion
    public void testVirtualhost() {
```
Typo.

s3/src/test/java/ch/cyberduck/core/s3/S3MoveFeatureTest.java
```diff
@@ -60,6 +60,18 @@ public void testMove() throws Exception {
         new S3DefaultDeleteFeature(session).delete(Collections.singletonList(renamed), new DisabledLoginCallback(), new Delete.DisabledCallback());
     }
 
+    @Test
+    public void testMoveVirtualHost() throws Exception {
```
AliveDevil, line 64: Add tests for folders as well.

AliveDevil (changes requested):

AliveDevil:
Adding to the "not working":
It doesn't work with the general S3 profile.
Creating a separate profile with a default hostname works, though.

AliveDevil (approved):

PR #12882, opened 2022-02-14
Fix #12880. Enable authentication using OpenSSH agent on Windows.

dkocher:

PR #12895, opened 2022-02-16
Implement the wrapper around named pipes

### LINKED ISSUE #12880: OpenSSH ssh-agent support for windows
Managing SSH hosts, their configuration, including keys, are usually done through .ssh/config. Recent versions of Windows (10+) come with a OpenSSH installation, including the `ssh-agent`. The configuration & keys loaded by the SSH agent do not seem to work:
* CyberDuck is unable to resolve hostnames listed in the .ssh/config file.
* Based on the documentation, `ssh-agent` is only supported on Linux (and Mac?)

**Describe the solution you'd like**
Currently Cyberduck (and Mountain Duck) seem to only support PuTTY & PageAnt, OpenSSH support on Windows would be much appreciated.

**Describe alternatives you've considered**
* Falling back to using PuTTY & PageAnt, requires me to maintain multiple programs for the same task.

**Additional context**

8uurg:
Adds an implementation for a socket based off the general idea described in ymnk/jsch-agent-proxy#34

Should remedy the issue faced in #12882 regarding named pipes, if the issue below can be fixed.

This general approach seems to work when running similar code using the
JVM, yet seems to fail when used in conjunction with Cyberduck:
```
java.io.FileNotFoundException: 'FileStream was asked to open a device that was not a file. For support for devices like 'com1:' or 'lpt1:', call CreateFile, then use the FileStream constructors that take an OS handle as an IntPtr.'

This exception was originally thrown at this call stack:
    java.io.FileDescriptor.open(string, int, int)
    java.io.FileDescriptor.openReadWrite(string)
    java.io.RandomAccessFile.open(string, int)
    java.io.RandomAccessFile.RandomAccessFile(java.io.File, string)
    java.io.RandomAccessFile.RandomAccessFile(string, string)
    ch.cyberduck.core.sftp.openssh.RandomAccessFileSocketFactory.WindowsSocket.WindowsSocket(ch.cyberduck.core.sftp.openssh.RandomAccessFileSocketFactory, string) in OpenSSHAgentAuthenticator.java
    ch.cyberduck.core.sftp.openssh.RandomAccessFileSocketFactory.open(string) in OpenSSHAgentAuthenticator.java
    com.jcraft.jsch.agentproxy.connector.SSHAgentConnector.open() in SSHAgentConnector.java
    com.jcraft.jsch.agentproxy.connector.SSHAgentConnector.SSHAgentConnector(com.jcraft.jsch.agentproxy.USocketFactory, string) in SSHAgentConnector.java
```
Some Googling indicates that this likely originates from C#,
so an alternate approach -- or a fix on the C# side -- may be necessary.

### DIFF
--- a/ssh/src/main/java/ch/cyberduck/core/sftp/SFTPSession.java
+++ b/ssh/src/main/java/ch/cyberduck/core/sftp/SFTPSession.java
@@ -286,10 +286,13 @@ public class SFTPSession extends Session<SSHClient> {
                 switch(Factory.Platform.getDefault()) {
                     case windows:
                         defaultMethods.add(new SFTPAgentAuthentication(client, new PageantAuthenticator()));
-                        // Break through
+                        
+                        defaultMethods.add(new SFTPAgentAuthentication(client, new OpenSSHAgentAuthenticator(
+                                new OpenSSHIdentityAgentConfigurator().getIdentityAgent(host.getHostname()), true)));
+                        break;
                     default:
                         defaultMethods.add(new SFTPAgentAuthentication(client, new OpenSSHAgentAuthenticator(
-                                new OpenSSHIdentityAgentConfigurator().getIdentityAgent(host.getHostname()))));
+                                new OpenSSHIdentityAgentConfigurator().getIdentityAgent(host.getHostname()), false)));
                         break;
                 }
             }
--- a/ssh/src/main/java/ch/cyberduck/core/sftp/openssh/OpenSSHAgentAuthenticator.java
+++ b/ssh/src/main/java/ch/cyberduck/core/sftp/openssh/OpenSSHAgentAuthenticator.java
@@ -23,6 +23,9 @@ import ch.cyberduck.core.sftp.auth.AgentAuthenticator;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
+import java.io.RandomAccessFile;
+import java.io.EOFException;
+import java.io.IOException;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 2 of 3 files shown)


AliveDevil:
Thanks for your effort!

On Windows we are using IKVM to transpile Java Bytecode into MSIL, therein lies the error you are seeing.
I‘ll check how this works for IKVM, to fix this.

AliveDevil:
A quick test revealed:
On .NET (Core, 5, 6) this works: `File.Open(@"\\.\pipe\openssh-ssh-agent", FileMode.Open, FileAccess.ReadWrite)`.
On .NET Framework (our target) this does not work, and fails with the same exception you got.

Though I managed to find _all_ references to named pipes here:
```
ikvm\jdk8u312-b07\hotspot\src\os\windows\vm\os_windows.cpp:
 4617    }
 4618  
 4619:   if (! ::PeekNamedPipe(han, NULL, 0, NULL, (LPDWORD)pbytes, NULL)) {
 4620:         /* PeekNamedPipe fails when at EOF.  In that case we
 4621           * simply make *pbytes = 0 which is consistent with the
 4622           * behavior we get on Solaris when an fd is at EOF.

ikvm\jdk8u312-b07\jdk\src\windows\native\java\io\io_util_md.c:
  340      }
  341  
  342:     if (! PeekNamedPipe(han, NULL, 0, NULL, pbytes, NULL)) {
  343:         /* PeekNamedPipe fails when at EOF.  In that case we
  344           * simply make *pbytes = 0 which is consistent with the
  345           * behavior we get on Solaris when an fd is at EOF.

ikvm\jdk8u312-b07\jdk\src\windows\native\sun\tools\attach\WindowsVirtualMachine.c:
  264      // Custom Security Descriptor is required here to "get" Medium Integrity Level.
  265      // In order to allow Medium Integrity Level clients to open
  266:     // and use a NamedPipe created by an High Integrity Level process.
  267      TCHAR *szSD = TEXT("D:")                  // Discretionary ACL
  268                    TEXT("(A;OICI;GRGW;;;WD)")  // Allow read/write to Everybody
  ...
  281      jstring_to_cstring(env, pipename, name, MAX_PIPE_NAME_LENGTH);
  282  
  283:     hPipe = CreateNamedPipe(
  284            name,                         // pipe name
  285            PIPE_ACCESS_INBOUND,          // read access
  ...
  296  
  297      if (hPipe == INVALID_HANDLE_VALUE) {
  298:         JNU_ThrowIOExceptionWithLastError(env, "CreateNamedPipe failed");
  299      }
  300      return (jlong)hPipe;
  ...
  322      BOOL fConnected;
  323  
  324:     fConnected = ConnectNamedPipe((HANDLE)hPipe, NULL) ?
  325          TRUE : (GetLastError() == ERROR_PIPE_CONNECTED);
  326      if (!fConnected) {
  327:         JNU_ThrowIOExceptionWithLastError(env, "ConnectNamedPipe failed");
  328      }
  329  }
  ```
  
  It looks like we are out of luck here and need to use JNA to access named pipes, instead of RandomReadWrite file accessors.

AliveDevil:
Culprit lies here: https://github.com/iterate-ch/ikvm/blob/master/ikvm/runtime/openjdk/java.io.cs#L126-L141
Either we need to extend the logic there to redirect `\\.\pipe` to `NamedPipeStream` … or completely remove all managed code and rely on PInvoke-calls for `CreateFile`[^1], which handles named pipes transparently.
And adapt our logic to JDK sources[^2].


[^1]: https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilew
[^2]: https://hg.openjdk.java.net/jdk8u/jdk8u-dev/jdk/file/7fcf35286d52/src/windows/native/java/io/io_util_md.c#l227

8uurg:
Going off the reference sources, the exception is being thrown at a code path only for a FileStream with a string path as input. Obtaining a handle ourselves (like the OpenJDK sources) and wrapping it in an owned SafeFileHandle[^1] could work as well.
But given the amount of security checks that the string based interface is doing, I am not entirely sure this would be OK.

Comparatively, the NamedPipeStream approach, while it seems hackier, is unlikely to affect other code paths, and does not require any interfacing with native libraries. 

In any case, my feature request has become quite a bit more complicated than I was initially anticipating.

[^1]: https://docs.microsoft.com/en-us/dotnet/api/microsoft.win32.safehandles.safefilehandle?view=netframework-4.8

AliveDevil:
Sorry for merging, Git is sometimes really stupid.
However, would you mind reopening the PR and applying these patches?
[GH-12895.zip](https://github.com/iterate-ch/cyberduck/files/8151599/GH-12895.zip)

Then sign the CLA and we can proceed to update the OpenSSH-Agent implementation on Windows.

Issue #12805, opened 2022-01-20
Aggregate for Cyberduck Windows native refresh


AliveDevil:
* [ ] #12803
* [ ] #12804
  * [ ] #12802

Tracking effort to track how far Cyberduck modernization went.

AliveDevil:
Dropping support for Windows 7, 8, 8.1, 10 (older than 1607 / 14393), Server 2012 and Server 2012R2 once merged.[^1]

New minimum: Windows 10 Build 14393 (Windows Server 2016).

[^1]: https://github.com/iterate-ch/cyberduck/pull/13030#issuecomment-1100302379

PR #12918, opened 2022-02-24
 Modernize PInvoke with CSWin32

### LINKED ISSUE #12803: Refresh usage of Windows Credential manager to properly store credentials
Current implementation lacks compatibility and is prone to errors due to length limitations of password blobs in Windows Credential Manager.
Review implementation to include "attributes"/pages for increased payload size of up to 64 KiB - should be enough for a password.

### LINKED ISSUE #12804: Modernize PInvoke with CSWin32
Replace all custom PInvoke signatures etc. with properly generated methods from https://github.com/microsoft/cswin32.

AliveDevil:
First part of #12805.

Fixes #12803 
Fixes #12804


core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
```diff
@@ -49,30 +48,178 @@ public override void addPassword(Scheme scheme, int port, string hostName, strin
...
+                if (credential.getOauth().getExpiryInMilliseconds() != null)
+                {
+                    winCred.Attributes["OAuth Expiry"] = credential.getOauth().getExpiryInMilliseconds().longValue().ToString();
+                }
+            }
+            if (protocol.isPrivateKeyConfigurable())
+            {
+                winCred.Attributes["Private Key Passphrase"] = credential.getIdentityPassphrase();
+            }
+            if (credential.isPasswordAuthentication())
```
AliveDevil, line 184: Need to check here.
If password authentication is required, then if either username or password is empty the whole credential is not saved.

Can private key configurable be used with password authentication, and username is empty?

core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
```diff
@@ -49,30 +48,178 @@ public override void addPassword(Scheme scheme, int port, string hostName, strin
...
+            if (protocol.isHostnameConfigurable() || !(protocol.isTokenConfigurable() || protocol.isOAuthConfigurable()))
+            {
+                pathBuilder.Append(":" + bookmark.getHostname());
+                if (protocol.isPortConfigurable() && !Equals(protocol.getDefaultPort(), bookmark.getPort()))
+                {
+                    pathBuilder.Append(":" + bookmark.getPort());
+                }
+            }
+            targetBuilder.Path = pathBuilder.ToString();
+            targetBuilder.Query = "user=" + credentials.getUsername();
```
AliveDevil, line 220: Uses format `application.container.name:protocol.identifier(:port)?\?user=username`.
Should this be changed/updated?

AliveDevil:
Tests won't compile, need to fix that.

core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
```diff
@@ -40,7 +39,7 @@ public override void addPassword(string serviceName, string user, string passwor
             var hostUrl = $"{serviceName} - {user}";
             if (!WinCredentialManager.SaveCredentials(hostUrl, new NetworkCredential(user, password)))
             {
-                throw new LocalAccessDeniedException($"Could not save credentials for \"{hostUrl}\" to Windows Credential Manager.");
+                logger.error($"Could not save credentials for \"{hostUrl}\" to Windows Credential Manager.");
```
dkocher, line 42: Why no longer throwing failure?

core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
```diff
@@ -17,15 +17,14 @@
...
-using System.Collections.Generic;
-using System.Linq;
 using System.Net;
 using System.Text;
-using System.Threading.Tasks;
+using static Windows.Win32.Security.Credentials.CRED_PERSIST;
+using static Windows.Win32.Security.Credentials.CRED_TYPE;
 
 namespace Ch.Cyberduck.Core
 {
```
dkocher, line 30: Should probably just implement `HostPasswordStore` if overriding all implementations from `DefaultHostPasswordStore` anyways.

core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
dkocher: Should match logic in `DefaultHostPasswordStore` to make sure we have no regressions.

core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
AliveDevil: Is there a difference in throwing? Will be written into error log regardless, and prevents unecessary stack rewinds and exception handling.

core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
AliveDevil: I'd agree, unless there is logic in DefaultHostPasswordStore, that I don't want to replicate _again_. Thats why I overwrote the HostPasswordStore-implementation for the revised implementation, and falling back to DefaultHostPasswordStore-logic.
See original logic for CredentialManagerPasswordStore: https://github.com/iterate-ch/cyberduck/blob/master/core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
I didn't want to have three different proxies in https://github.com/iterate-ch/cyberduck/blob/feature/GH-12805-1/core/src/main/csharp/ch/cyberduck/core/PasswordStoreFacade.cs.

core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
AliveDevil: DefaultHostPasswordStore runs separate addPassword-calls for these conditions.

CredentialManagerPasswordStore now only performs one single commit with all fields combined.

AliveDevil:
Works For Me (tm)
```
[INFO]   [msbuild] C:\Program Files\Microsoft Visual Studio\2022\Enterprise\MSBuild\Microsoft\VC\v170\Microsoft.CppBuild.targets(437,5): error MSB8013: This project doesn't contain the Configuration and Platform combination of Release|x64. [D:\a\cyberduck\cyberduck\windows\src\main\wix\WindowsVersionCustomAction\WindowsVersionCustomAction.vcxproj]
```

tatze96:
[637825876270682181.txt](https://github.com/iterate-ch/cyberduck/files/8230295/637825876270682181.txt)
Cyberduck crasht beim verwenden des Disconnect Buttons.

AliveDevil:
Wouldn't merge for 8.3.

core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
AliveDevil: Upping this one.

`application.container.name` for Cyberduck is `duck`, for Mountain Duck Windows this is `Cyberduck`. This way Cyberduck and Mountain Duck don't share credentials, which is kind of unfortunate. Which key should be used for these shared elements? Taking into account branded builds.

dkocher (approved):

ylangisc (approved):

PR #12933, opened 2022-03-02
Add InputStream implementation that reads from a proxy plaintext stre…

ylangisc:
…am and encrypts the content.


dkocher:
Dependent on #12934 to reduce temporary disk usage.

dkocher (approved):

PR #12939, opened 2022-03-06
ssh config parsing: expand ~ in IdentityAgent path

### LINKED ISSUE #12954: Resolve tilde in IdentityAgent configuration

rufoa:
Currently if a user has an option in their ssh config file like this:
`IdentityAgent ~/.1password/agent.sock`
then cyberduck can't open the socket because the ~ in the path isn't expanded, it's treated as a literal ~

(But `IdentityAgent /Users/myname/.1password/agent.sock` works correctly)

This PR applies the same logic to `IdentityAgent` as is already used for `IdentityFile`, so paths with ~ work correctly.

Thanks

### DIFF
--- a/ssh/src/main/java/ch/cyberduck/core/sftp/openssh/OpenSSHIdentityAgentConfigurator.java
+++ b/ssh/src/main/java/ch/cyberduck/core/sftp/openssh/OpenSSHIdentityAgentConfigurator.java
@@ -18,6 +18,7 @@ package ch.cyberduck.core.sftp.openssh;
  * dkocher@cyberduck.ch
  */
 
+import ch.cyberduck.core.Local;
 import ch.cyberduck.core.LocalFactory;
 import ch.cyberduck.core.sftp.openssh.config.transport.OpenSshConfig;
 
@@ -33,7 +34,11 @@ public class OpenSSHIdentityAgentConfigurator {
     }
 
     public String getIdentityAgent(final String alias) {
-        return configuration.lookup(alias).getIdentityAgent();
+        final Local agent = configuration.lookup(alias).getIdentityAgent();
+        if(null == agent) {
+            return null;
+        }
+        return agent.getAbsolute();
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


dkocher (approved):

ssh/src/main/java/ch/cyberduck/core/sftp/openssh/config/transport/OpenSshConfig.java
```diff
@@ -230,7 +230,7 @@ else if("IdentityFile".equalsIgnoreCase(keyword)) {
             else if("IdentityAgent".equalsIgnoreCase(keyword)) {
                 for(final Host c : current) {
                     if(c.identityAgent == null) {
-                        c.identityAgent = dequote(argValue);
+                        c.identityAgent = LocalFactory.get(dequote(argValue));
```
ylangisc, line 233: `Local` to `String`.

dkocher (changes requested):
Type of `identityAgent` must be changed accordingly.

dkocher (changes requested):
Review usages of `getIdentityAgent`.

rufoa:
Is this OK or would you rather move the `.getAbbreviatedPath()` call to here instead?

https://github.com/iterate-ch/cyberduck/blob/4ea8d5d6654dc0be43bd6407a5b5e9b0fc63241b/ssh/src/main/java/ch/cyberduck/core/sftp/SFTPSession.java#L301

ssh/src/main/java/ch/cyberduck/core/sftp/openssh/OpenSSHIdentityAgentConfigurator.java
```diff
@@ -33,7 +33,7 @@ public OpenSSHIdentityAgentConfigurator(final OpenSshConfig configuration) {
     }
 
     public String getIdentityAgent(final String alias) {
-        return configuration.lookup(alias).getIdentityAgent();
+        return configuration.lookup(alias).getIdentityAgent().getAbbreviatedPath();
```
dkocher, line 36: This will again lead to the same problem if you return the abbreviated path here. Thus you should call `getAbsolute()` I guess.

dkocher (changes requested):

ssh/src/main/java/ch/cyberduck/core/sftp/openssh/OpenSSHIdentityAgentConfigurator.java
rufoa: right. Looks like we also need to handle the null case

dkocher (changes requested):
Can you please sqash the commits to a single commit using `rebase -i master`.

dkocher (approved):

PR #12979, opened 2022-03-17
ConnectionTimeout service handling min and max values

AliveDevil:
Cannot set no maximum, cannot set no minimum, but throws ArgumentOutOfRange if value is outside [Min,Max]
This forces connection.timeout.seconds to be within [Min,Max] on load of default.properties, potentially ignoring user-defined values outside of range in default.properties.

### DIFF
--- a/core/src/main/java/ch/cyberduck/core/DefaultConnectionTimeout.java
+++ b/core/src/main/java/ch/cyberduck/core/DefaultConnectionTimeout.java
@@ -23,23 +23,21 @@ import ch.cyberduck.core.preferences.PreferencesReader;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
-public class DefaultConnectionTimeout implements ConnectionTimeout.MutableConnectionTimeout {
+public class DefaultConnectionTimeout implements ConnectionTimeout {
     private static final Logger log = LogManager.getLogger(DefaultConnectionTimeout.class);
+    private static final Preferences mutablePreferences = PreferencesFactory.get();
 
     public final static int TIMEOUT_MIN = 10;
     public final static int TIMEOUT_MAX = 60;
 
-    private final Preferences mutablePreferences;
     private final PreferencesReader preferences;
 
     public DefaultConnectionTimeout() {
-        mutablePreferences = PreferencesFactory.get();
-        preferences = mutablePreferences;
+        this(mutablePreferences);
     }
 
     public DefaultConnectionTimeout(final PreferencesReader preferences) {
         this.preferences = preferences;
-        mutablePreferences = null;
     }
 
     @Override
@@ -49,10 +47,6 @@ public class DefaultConnectionTimeout implements ConnectionTimeout.MutableConnec
 
     @Override
     public void setTimeout(final int timeout) {
-        if(mutablePreferences == null) {
-            log.warn(String.format("Trying to set %s to %d on Host preferences.", PREFERENCE_KEY, timeout));
-            return;
-        }
         mutablePreferences.setProperty(PREFERENCE_KEY, clamp(timeout));
     }
 
--- a/core/src/main/java/ch/cyberduck/core/DefaultConnectionTimeout.java
+++ b/core/src/main/java/ch/cyberduck/core/DefaultConnectionTimeout.java
@@ -1,0 +1,68 @@
+package ch.cyberduck.core;
+
+/*
+ * Copyright (c) 2002-2022 iterate GmbH. All rights reserved.
+ * https://cyberduck.io/
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ */
+
+import ch.cyberduck.core.preferences.HostPreferences;
+import ch.cyberduck.core.preferences.Preferences;
+import ch.cyberduck.core.preferences.PreferencesFactory;
+import ch.cyberduck.core.preferences.PreferencesReader;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+public class DefaultConnectionTimeout implements ConnectionTimeout.MutableConnectionTimeout {
+    private static final Logger log = LogManager.getLogger(DefaultConnectionTimeout.class);
+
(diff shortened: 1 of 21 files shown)


ylangisc (approved):

dkocher (changes requested):
Can this instead be moved to where this is used in Preferences adjusting to min/max values accordingly.

AliveDevil:
> Can this instead be moved to where this is used in Preferences adjusting to min/max values accordingly.

Then non-clamped defaults are applied until the user opens the preferences UI for the first time, which then clamps user default setting to user preference into min/max.

Have timeout settings be applied partially (e.g. 120 seconds timeout works for user), user opens preferences, resets to 60, which won't work for user again. It should be as consistent as possible, not dependent on whether user opened preferences or not.

dkocher:
> It should be as consistent as possible, not dependent on whether user opened preferences or not.

Then I propose we add a service similar to `TransferConnectionLimiter` that handles all cases or we add validation to preferences in a generic way.

AliveDevil:
> service similar to `TransferConnectionLimiter`

Which implements what I did in the app startup for Windows?
This service should clamp the timeout. Min/Max configurable through concrete implementation?

> validation to preferences in a generic way

Sounds good to me. Though I'm not sure how we'd proceed here. Service looks to be easier to implement.

AliveDevil:
Proceeding to implement the timeout service with auto clamping.

dracoon/src/main/java/ch/cyberduck/core/sds/SDSSession.java
```diff
@@ -256,7 +256,7 @@ public void process(final HttpRequest request, final HttpContext context) {
                 .register(new JSON())
                 .register(JacksonFeature.class)
                 .connectorProvider(new HttpComponentsProvider(apache))));
-        final int timeout = preferences.getInteger("connection.timeout.seconds") * 1000;
+        final int timeout = ConnectionTimeoutFactory.get().getTimeout() * 1000;
```
AliveDevil, line 259: Should HostPreferences be used here?

ftp/src/main/java/ch/cyberduck/core/ftp/FTPSession.java
```diff
@@ -139,7 +140,7 @@ protected void configure(final FTPClient client) throws IOException {
         client.setProtocol(host.getProtocol());
         client.setSocketFactory(new ProxySocketFactory(host));
         client.setControlEncoding(host.getEncoding());
-        final int timeout = preferences.getInteger("connection.timeout.seconds") * 1000;
+        final int timeout = ConnectionTimeoutFactory.get().getTimeout() * 1000;
```
AliveDevil, line 143: Should HostPreferences be used here?

ssh/src/main/java/ch/cyberduck/core/sftp/SFTPSession.java
```diff
@@ -183,7 +183,7 @@ protected SSHClient connect(final HostKeyCallback key, final LoginCallback promp
 
     private SSHClient toClient(final HostKeyCallback key, final Config configuration) {
         final SSHClient connection = new SSHClient(configuration);
-        final int timeout = preferences.getInteger("connection.timeout.seconds") * 1000;
+        final int timeout = ConnectionTimeoutFactory.get(preferences).getTimeout() * 1000;
```
AliveDevil, line 186: Should HostPreferences be used here?

ssh/src/main/java/ch/cyberduck/core/sftp/SFTPSession.java
```diff
@@ -261,7 +261,7 @@ public void login(final Proxy proxy, final LoginCallback prompt, final CancelCal
         this.authenticate(client, host, prompt, cancel);
         try {
             sftp = new LoggingSFTPEngine(client, this).init();
-            sftp.setTimeoutMs(preferences.getInteger("connection.timeout.seconds") * 1000);
+            sftp.setTimeoutMs(ConnectionTimeoutFactory.get(preferences).getTimeout() * 1000);
```
AliveDevil, line 264: Should HostPreferences be used here?

storegate/src/main/java/ch/cyberduck/core/storegate/StoregateSession.java
```diff
@@ -124,7 +125,7 @@ public void process(final HttpRequest request, final HttpContext context) {
             .register(new JSON())
             .register(JacksonFeature.class)
             .connectorProvider(new HttpComponentsProvider(apache))));
-        final int timeout = preferences.getInteger("connection.timeout.seconds") * 1000;
+        final int timeout = ConnectionTimeoutFactory.get().getTimeout() * 1000;
```
AliveDevil, line 128: Should HostPreferences be used here?

dracoon/src/main/java/ch/cyberduck/core/sds/SDSSession.java
dkocher: We can by default refactor this to `HostPreferences` instead of directly accessing Preferences.

ftp/src/main/java/ch/cyberduck/core/ftp/FTPSession.java
AliveDevil: Didn't see preferences already is HostPreferences-instance.

dracoon/src/main/java/ch/cyberduck/core/sds/SDSSession.java
AliveDevil: Didn't see preferences already is HostPreferences-instance.

storegate/src/main/java/ch/cyberduck/core/storegate/StoregateSession.java
AliveDevil: Changed to get ConnectionTimeout for Host, instead of global preferences.

core/src/main/java/ch/cyberduck/core/DefaultConnectionTimeout.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    @Override
+    public int getTimeout() {
+        return clamp(preferences.getInteger(PREFERENCE_KEY));
+    }
+
+    @Override
+    public void setTimeout(final int timeout) {
+        if(mutablePreferences == null) {
```
dkocher, line 52: Can this be simplified that setting the timeout will always use `PreferencesFactory.get()` to remove the checks and overloaded constructor?

dkocher (changes requested):

core/src/main/java/ch/cyberduck/core/DefaultConnectionTimeout.java
AliveDevil: This still requires an overloaded constructor, due to UI not having a Host configured, and passing `null` to `HostPreferences` won't work for reading preference values.

dkocher (approved):

PR #13000, opened 2022-03-22
Changed copy() implementation to channelCopy(), with a 100% speed increase

### LINKED ISSUE #12996: Compose segments of download very slow after downloading is slow
After downloading a multi-segment download of large size (200GB etc.) the recompositing of the file is extremely slow, like hours, longer than the download. I see this was closed previously with a 'don't use segments' workaround, but doing so trashes your ability to 
- get maximum download speed
- effectively resume download, by resuming missing segments

I don't think the workaround solves the problem. 

I'd be happy to contribute a fix for this, if someone can point me towards the area of the codebase that functionally recreates the downloaded file from the multiple chunks.

Thanks!

yardik:
- about 2x faster than the byte stream implementation
- all existing tests pass
- added new test for aggregation

### DIFF
--- a/core/src/main/java/ch/cyberduck/core/Local.java
+++ b/core/src/main/java/ch/cyberduck/core/Local.java
@@ -370,12 +370,12 @@ public class Local extends AbstractPath implements Referenceable, Serializable {
             if(log.isDebugEnabled()) {
                 log.debug(String.format("Copy to %s with options %s", copy, options));
             }
-            InputStream in = null;
-            OutputStream out = null;
+            FileChannel in = null;
+            FileChannel out = null;
             try {
-                in = this.getInputStream();
-                out = copy.getOutputStream(options.append);
-                IOUtils.copy(in, out);
+                in = this.getReadChannel(this.path);
+                out = copy.getWriteChannel(copy.path, options.append);
+                out.transferFrom(in, out.size(), in.size());
             }
             catch(IOException e) {
                 throw new LocalAccessDeniedException(MessageFormat.format(
@@ -426,25 +426,38 @@ public class Local extends AbstractPath implements Referenceable, Serializable {
         return String.format("file:%s", path);
     }
 
+
     public InputStream getInputStream() throws AccessDeniedException {
-        return this.getInputStream(path);
+        return getInputStream(path);
     }
 
-    protected InputStream getInputStream(final String path) throws LocalAccessDeniedException {
+    protected InputStream getInputStream(final String path) throws AccessDeniedException {
         try {
-            final FileChannel channel = FileChannel.open(Paths.get(path), StandardOpenOption.READ);
-            return new SeekableByteChannelInputStream(channel);
+            return new SeekableByteChannelInputStream(getReadChannel(path));
+        }
+        catch(RuntimeException e) {
+            throw new LocalAccessDeniedException(e.getMessage(), e);
+        }
+    }
+
+    private FileChannel getReadChannel(final String path) throws LocalAccessDeniedException {
+        try {
+            return FileChannel.open(Paths.get(path), StandardOpenOption.READ);
         }
         catch(RuntimeException | IOException e) {
             throw new LocalAccessDeniedException(e.getMessage(), e);
         }
     }
 
-    public OutputStream getOutputStream(final boolean append) throws AccessDeniedException {
-        return this.getOutputStream(path, append);
+    protected OutputStream getOutputStream(final String path, final boolean append) throws AccessDeniedException {
+        return Channels.newOutputStream(getWriteChannel(path, append));
     }
 
-    protected OutputStream getOutputStream(final String path, final boolean append) throws LocalAccessDeniedException {
+    public OutputStream getOutputStream(final boolean append) throws AccessDeniedException {
+        return Channels.newOutputStream(getWriteChannel(path, append));
+    }
+
+    private FileChannel getWriteChannel(final String path, final boolean append) throws LocalAccessDeniedException {
         try {
             final Set<OpenOption> options = new HashSet<>();
             options.add(StandardOpenOption.WRITE);
@@ -457,8 +470,7 @@ public class Local extends AbstractPath implements Referenceable, Serializable {
             else {
                 options.add(StandardOpenOption.TRUNCATE_EXISTING);
             }
-            final FileChannel channel = FileChannel.open(Paths.get(path), options);
-            return Channels.newOutputStream(channel);
+            return FileChannel.open(Paths.get(path), options);
         }
         catch(RuntimeException | IOException e) {
             throw new LocalAccessDeniedException(e.getMessage(), e);
--- a/core/src/test/java/ch/cyberduck/core/local/LocalTest.java
+++ b/core/src/test/java/ch/cyberduck/core/local/LocalTest.java
@@ -7,9 +7,16 @@ import ch.cyberduck.core.exception.AccessDeniedException;
 import ch.cyberduck.core.exception.LocalAccessDeniedException;
 import ch.cyberduck.core.preferences.PreferencesFactory;
 
+import org.apache.commons.io.FileUtils;
+import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.junit.Test;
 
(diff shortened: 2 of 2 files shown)


yardik:
@dkocher that PR we discussed in #12996

core/src/main/java/ch/cyberduck/core/Local.java
```diff
@@ -359,7 +359,33 @@ public void rename(final Local renamed) throws AccessDeniedException {
...
+                    LocaleFactory.localizedString("Cannot copy {0}", "Error"), this.getName()), e);
+            }
+            finally {
+                IOUtils.closeQuietly(c1);
+                IOUtils.closeQuietly(c2);
+            }
+        }
     }
 
     public void copy(final Local copy, final CopyOptions options) throws AccessDeniedException {
```
dkocher, line 391: I see no reason to keep this old implementation but would just replace it with the new implementation using the `FileChannel`.

core/src/main/java/ch/cyberduck/core/Local.java
```diff
@@ -444,6 +479,27 @@ public OutputStream getOutputStream(final boolean append) throws AccessDeniedExc
         return this.getOutputStream(path, append);
     }
 
+    protected FileChannel getRawChannelOutputStream(final boolean append) throws LocalAccessDeniedException {
+        try {
+            final Set<OpenOption> options = new HashSet<>();
```
dkocher, line 484: Can we extract a method to open the channel to remove the duplicated code.

core/src/main/java/ch/cyberduck/core/transfer/download/AbstractDownloadFilter.java
```diff
@@ -284,7 +284,7 @@ public void complete(final Path file, final Local local,
                     if(log.isInfoEnabled()) {
                         log.info(String.format("Append segment %s to %s", segmentFile, local));
                     }
-                    segmentFile.copy(local, new Local.CopyOptions().append(true));
+                    segmentFile.channelCopy(local, new Local.CopyOptions().append(true));
```
dkocher, line 287: Revert to `copy` when previous implementation is removed and new implementation takes previous signature.

core/src/main/java/ch/cyberduck/core/Local.java
```diff
@@ -444,6 +479,27 @@ public OutputStream getOutputStream(final boolean append) throws AccessDeniedExc
...
+            if(!this.exists()) {
+                options.add(StandardOpenOption.CREATE);
+            }
+            if(append) {
+                options.add(StandardOpenOption.APPEND);
+            }
+            else {
+                options.add(StandardOpenOption.TRUNCATE_EXISTING);
+            }
+            final FileChannel channel = FileChannel.open(Paths.get(path), options);
```
dkocher, line 495: Inline return value.

dkocher (changes requested):

core/src/main/java/ch/cyberduck/core/Local.java
yardik: fixed

core/src/main/java/ch/cyberduck/core/Local.java
yardik: fixed

core/src/main/java/ch/cyberduck/core/transfer/download/AbstractDownloadFilter.java
yardik: fixed

core/src/main/java/ch/cyberduck/core/Local.java
yardik: fixed

yardik:
Updated as per all requests. Thanks for the feedback!

core/src/main/java/ch/cyberduck/core/Local.java
```diff
@@ -370,20 +370,20 @@ public void copy(final Local copy, final CopyOptions options) throws AccessDenie
             if(log.isDebugEnabled()) {
                 log.debug(String.format("Copy to %s with options %s", copy, options));
             }
-            InputStream in = null;
-            OutputStream out = null;
+            FileChannel c1 = null;
```
dkocher, line 373: Could you rename the local variable back to `in` and `out` respectively to make it clear what the source and target of the copy operation is.

core/src/main/java/ch/cyberduck/core/Local.java
```diff
@@ -426,25 +426,38 @@ public String toURL() {
...
     }
 
+
     public InputStream getInputStream() throws AccessDeniedException {
-        return this.getInputStream(path);
+        return getInputStream(path);
     }
 
-    protected InputStream getInputStream(final String path) throws LocalAccessDeniedException {
+    public InputStream getInputStream(final String path) throws AccessDeniedException {
```
dkocher, line 434: This can remain `protected` or even `private` I guess.

core/src/main/java/ch/cyberduck/core/Local.java
```diff
@@ -426,25 +426,38 @@ public String toURL() {
...
-            final FileChannel channel = FileChannel.open(Paths.get(path), StandardOpenOption.READ);
-            return new SeekableByteChannelInputStream(channel);
+            return new SeekableByteChannelInputStream(getRawChannelInputStream(path));
+        }
+        catch(RuntimeException e) {
+            throw new LocalAccessDeniedException(e.getMessage(), e);
+        }
+    }
+
+    public FileChannel getRawChannelInputStream(final String path) throws LocalAccessDeniedException {
```
dkocher, line 443: Can be changed to `private` scope possibly.

core/src/main/java/ch/cyberduck/core/Local.java
```diff
@@ -426,25 +426,38 @@ public String toURL() {
...
+        return Channels.newOutputStream(getRawChannelOutputStream(path, append));
+    }
+
     public OutputStream getOutputStream(final boolean append) throws AccessDeniedException {
-        return this.getOutputStream(path, append);
+        return Channels.newOutputStream(getRawChannelOutputStream(path, append));
     }
 
-    protected OutputStream getOutputStream(final String path, final boolean append) throws LocalAccessDeniedException {
+    protected FileChannel getRawChannelOutputStream(final String path, final boolean append) throws LocalAccessDeniedException {
```
dkocher, line 460: Can be changed to `private` scope possibly.

dkocher (changes requested):

core/src/main/java/ch/cyberduck/core/Local.java
yardik: fixed

core/src/main/java/ch/cyberduck/core/Local.java
yardik: fixed

core/src/main/java/ch/cyberduck/core/Local.java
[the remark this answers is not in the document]
yardik: fixed

core/src/main/java/ch/cyberduck/core/Local.java
yardik: fixed

yardik:
Updated as requested, thanks!

core/src/main/java/ch/cyberduck/core/Local.java
```diff
@@ -426,25 +426,38 @@ public String toURL() {
...
-            final FileChannel channel = FileChannel.open(Paths.get(path), StandardOpenOption.READ);
-            return new SeekableByteChannelInputStream(channel);
+            return new SeekableByteChannelInputStream(getReadChannel(path));
+        }
+        catch(RuntimeException e) {
+            throw new LocalAccessDeniedException(e.getMessage(), e);
+        }
+    }
+
+    public FileChannel getReadChannel(final String path) throws LocalAccessDeniedException {
```
dkocher, line 443: Change modifier to `private` as well.

dkocher (changes requested):

yardik:
fixed

dkocher (approved):

PR #13299, opened 2022-04-16
Add profile for Yandex Object Storage

Heitzz:
Yandex Object Storage is a universal, scalable solution for data storage. It is the perfect choice both for high-load services that require reliable and fast access to data, and for projects with minimal requirements for storage infrastructure.

It also has a [wiki page](https://cloud.yandex.com/en/docs/storage/tools/cyberduck) for cyberduck.
It will be much more convenient if cyberduck has a profile for it.


dkocher:
We will not want to advertise Yandex with Russian state bank VTB sanctioned because of the Russian invasion in Ukraine having stakes in.

PR #13319, opened 2022-04-22
Make sure that querying S3 direct upload supported works in offline mode.

ylangisc:


ylangisc:
We should consider persisting backend metadata such as software version, s3 support, ... in our local cache.

dkocher (changes requested):
I would rather make sure settings are cached regardless if expired that we must not need nil checks. Should only be nullified when session is closed. Object holder could be rewritten to refresh when expired instead of nullifying.

dkocher (approved):

PR #13338, opened 2022-04-28
Fix regression from ddd7c79b which causes freeze on main thread when …

### LINKED ISSUE #13322: freeze when uploading multiple files
Cyberduck for macOS seems to lock up when uploading multiple files via FTP to certain servers, particularly those hosted on Doteasy.  Attempting to transfer multiple files via drag-and-drop to the browser window results in an endless progress bar, requiring a FORCE QUIT of the Cyberduck app.  Using Cyberduck 8.3.2 on macOS 12.3.1.  Have attempted this process for numerous different clients whose sites are hosted on Doteasy; it happens every time.

dkocher:
…waiting for pool shutdown on main thread while awaiting dismiss of error callback prompt from user.


AliveDevil (approved):
Confirmed on Windows, error messages are thrown correctly and report issues without locking up.

Issue #10560, opened 2018-12-17
Support versioning in NextCloud & ownCloud


cyberduck:
> **@dkocher** created the issue

> https://docs.nextcloud.com/server/14/developer_manual/client_apis/WebDAV/versions.html

dkocher:
As for other protocols in #13342.

dkocher:
Documentation for ownCloud

* [File Versions](https://doc.owncloud.com/server/next/developer_manual/webdav_api/files_versions.html)

PR #13342, opened 2022-04-29
Add feature to list previous versions to interface

dkocher:


dkocher:
Eventually also #10560.

AliveDevil:
Requires modification of https://github.com/iterate-ch/onedrive-java-client in order to support restore version[^1].

[^1]: https://docs.microsoft.com/en-us/graph/api/driveitemversion-restore?view=graph-rest-1.0&tabs=http

core/native/refresh/src/main/csharp/ViewModels/Info/VersionsViewModel.cs
```diff
@@ -0,0 +1,204 @@
...
+                                LoginCallbackFactory.get(controller), norm,
+                                PathCache.empty(), new DisabledProgressListener())));
+                    await result.Task;
+                }
+                finally
+                {
+                    Busy = false;
+                }
+                await Load.ExecuteIfPossible();
+            }, this.WhenAnyValue(v => v.SelectedVersionValue).Select(v => v != null));
```
AliveDevil, line 86: Shouldn't we check whether Delete is possible?
i.e. OneDrive cannot delete single versions.

ylangisc (approved):

Issue #13376, opened 2022-05-11
Add support for Cloudflare R2


ahrenstein:
**Is your feature request related to a problem? Please describe.**
Cloudflare is launching a new S3 compatible object storage service called [R2](https://blog.cloudflare.com/r2-open-beta/).
Cyberduck (and Mountain Duck) should definitely add support for this new service as it has a lot of benefits over AWS S3 in different cases. (One of which being zero cost on egress)

**Describe the solution you'd like**
I'd like to see support added for Cloudflare R2 (which is currently in open beta)

**Describe alternatives you've considered**
In the meantime, I'll see if I can create a custom profile to work with R2 but long term official support would be great.

**Additional context**
There is some [official documentation](https://developers.cloudflare.com/r2/) along with the most relevant doc being the [current API implementation status](https://developers.cloudflare.com/r2/platform/s3-compatibility/api/).  
Additionally the document about [using Wrangler to create and manage buckets](https://developers.cloudflare.com/r2/get-started/) might be important.

juliancox:
Using S3 mode I can connect to R2 and list buckets but attempting to list objects in a bucket gives the following error:
Listing directory gc-public failed.
Internal Server Error. Request Error [InternalError]. Please contact your web hosting service provider for assistance.

AliveDevil:
Can you check whether forcing path style requests works for R2? You can enable this profile in Cyberduck Preferences -> Profiles -> search for "S3 (Deprecated path style requests)"

There is a bunch of unsupported APIs which Cyberduck may depend on ([Unimplemented object-level operations](https://developers.cloudflare.com/r2/platform/s3-compatibility/api/#unimplemented-object-level-operations)) - specifically `ListObjects`. Chances are higher if there is a Java SDK or OpenAPI specification to be implemented, though.

Cloudflare isn't that great of making sure their blog posts are accurate, though.
![image](https://user-images.githubusercontent.com/734738/168018604-eb5770a1-e963-4af3-8e66-eb1f611a4668.png)
Referring in the text to "bucket.account.r2storage.com" while the anchor links to "account.r2.cloudflarestorage.com/bucket" isn't something I'd want to see.

juliancox:
@AliveDevil I can confirm S3 (Deprecated path style requests) profile seems to allow R2 connections to work as expected I can list buckets, download and upload files without any problems.  

Cloudflare docs do leave a lot to be desired - sigh!

deepybee:
[...]

PR #13385, opened 2022-05-13
Add Cloudflare R2 storage s3-compat profile

AliveDevil:
As part of #13376

Refer iterate-ch/docs#198

### DIFF
--- a/profiles/Cloudflare R2 Storage (S3).cyberduckprofile
+++ b/profiles/Cloudflare R2 Storage (S3).cyberduckprofile
@@ -1,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+    <dict>
+        <key>Protocol</key>
+        <string>s3</string>
+        <key>Vendor</key>
+        <string>s3-cloudflare-r2</string>
+        <key>Scheme</key>
+        <string>https</string>
+        <key>Description</key>
+        <string>Cloudflare R2 Storage (S3)</string>
+        <key>Default Hostname</key>
+        <string>&lt;accountid&gt;.r2.cloudflarestorage.com</string>
+        <key>Hostname Configurable</key>
+        <true/>
+        <key>Port Configurable</key>
+        <false/>
+        <key>Region</key>
+        <string>auto</string>
+        <key>Regions</key>
+        <array>
+            <string>auto</string>
+        </array>
+        <key>Properties</key>
+        <array>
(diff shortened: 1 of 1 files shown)


profiles/Cloudflare R2 Storage (S3).cyberduckprofile
```diff
@@ -0,0 +1,24 @@
...
+        <string>Cloudflare R2 Storage (S3)</string>
+        <key>Default Hostname</key>
+        <string>accountid.r2.cloudflarestorage.com</string>
+        <key>Hostname Configurable</key>
+        <true/>
+        <key>Port Configurable</key>
+        <false/>
+        <key>Properties</key>
+        <array>
+            <string>s3service.disable-dns-buckets=true</string>
```
dkocher, line 21: - Add setting for storage classes available
- Add definition for default region with `<key>Region</key>` and available regions `<key>Region</key>`
- Add help for https://docs.cyberduck.io/protocols/s3/cloudflare

dkocher, line 21: - Add setting for storage classes available like `s3.storage.class.options=STANDARD GLACIER`
- Add definition for default region with `<key>Region</key>` and available regions `<key>Region</key>`
- Add help for https://docs.cyberduck.io/protocols/s3/cloudflare

AliveDevil: Unimplemented:
> CreateMultipartUpload:
> ❌ Storage class:
> ❌ x-amz-storage-class

Didn't find any evidence that there are regions. Docs are opaque here.

dkocher: Thus set 
- s3.storage.class.options=STANDARD
- `Region` and `Regions` explicitly to default to the value returned by the `?location` for buckets.

AliveDevil: >  by the ?location for buckets
Awaiting response.

dkocher:
Querying the bucket location is not supported

```
<Error><Code>NotImplemented</Code><Message>GetBucketLocation not implemented</Message></Error>
```

AliveDevil:
RClone is using "auto" Region. Using same value.

profiles/Cloudflare R2 Storage (S3).cyberduckprofile
AliveDevil: Using "auto" value of rclone: https://forum.rclone.org/t/cloudflare-r2-now-working-with-rclone/30730

dkocher (approved):

dkocher:
It may not be clear that `accountid` in the default hostname `accountid.r2.cloudflarestorage.com` needs to be replaced. I suppose it would be better to only set a placeholder now possible with #13396.

PR #13438, opened 2022-06-07
Remove workaround from 1103301e no longer required with server side fix.

dkocher:


freenet/src/main/java/ch/cyberduck/core/freenet/FreenetFindFeature.java
```diff
@@ -15,27 +15,12 @@
...
  */
 
-import ch.cyberduck.core.ListProgressListener;
-import ch.cyberduck.core.Path;
 import ch.cyberduck.core.dav.DAVFindFeature;
 import ch.cyberduck.core.dav.DAVSession;
-import ch.cyberduck.core.exception.BackgroundException;
-import ch.cyberduck.core.shared.DefaultFindFeature;
 
 public class FreenetFindFeature extends DAVFindFeature {
```
AliveDevil, line 21: Isn't this then redundant? If there is nothing left, except the constructor, then DAVFindFeature could be used instead of subclassing.

dkocher: I just keep it around for the integration tests.

AliveDevil (approved):

PR #13440, opened 2022-06-07
Support to delete files from Versions tab in Info panel

### LINKED ISSUE #13435: Permanently delete previous revisions of files

dkocher:


googledrive/src/main/java/ch/cyberduck/core/googledrive/DriveAttributesFinderFeature.java
```diff
@@ -104,7 +104,7 @@ public PathAttributes toAttributes(final File f) {
         if(null != f.getExplicitlyTrashed()) {
             if(f.getExplicitlyTrashed()) {
                 // Mark as hidden
-                attributes.setDuplicate(true);
+                attributes.setHidden(true);
```
ylangisc, line 107: Has been changed from `hidden` to `duplicate` and now `hidden` again - might lead to a regression?

dkocher: Has test coverage .

PR #13464, opened 2022-06-14
Add Infomaniak Public Cloud profile

kallioli:
Add connection profile for Infomaniak Public Cloud, Based on swift Object Storage.


dkocher (approved):
I have verified the connection profile.

PR #13538, opened 2022-07-04
Add alternate implementation for macOS 12.0+.

dkocher:


core/dylib/src/main/java/ch/cyberduck/core/preferences/ApplicationPreferences.java
```diff
@@ -84,7 +86,13 @@ protected void setFactories() {
         this.setDefault("factory.notification.class", NotificationCenter.class.getName());
         this.setDefault("factory.iconservice.class", WorkspaceIconService.class.getName());
         this.setDefault("factory.filedescriptor.class", LaunchServicesFileDescriptor.class.getName());
-        this.setDefault("factory.schemehandler.class", LaunchServicesSchemeHandler.class.getName());
+        if(Factory.Platform.osversion.matches("10\\.(12|13|14|15|16).*")) {
```
ylangisc, line 89: What about Big Sur (11.x)? It seems that it is correctly reported as of JDK 17, refer to https://bugs.openjdk.org/browse/JDK-8253702 and https://github.com/openjdk/jdk/commit/66757750

dkocher: Test the assumption in #13540.

ylangisc (changes requested):
Hanging tests on macOS 12.

dkocher:
> Hanging tests on macOS 12.

Unfortunately cannot be reproduced locally.

ylangisc:
> > Hanging tests on macOS 12.
> 
> Unfortunately cannot be reproduced locally.

Related to any headless setup of the CI env?

.github/workflows/build.yml
```diff
@@ -13,7 +13,7 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                os: [ macos-10.15, macos-11, macos-12, ubuntu-latest, windows-latest ]
+                os: [ macos-11, macos-12, ubuntu-latest, windows-latest ]
```
ylangisc, line 16: Why macOS 10.15 has been removed?

dkocher: > GitHub Actions The macOS 10.15 Actions runner image is being deprecated and will be removed by 8/30/22

ylangisc (approved):

PR #13546, opened 2022-07-06
Bring back support for default editor

### LINKED ISSUE #13541: Cannot choose a default editor in Windows
**Describe the bug**
When choosing a default editor it does not set properly, it goes through the actions of allowing you to choose an editor but it is never used.

**To Reproduce**
1. Click on **Edit**
2. Click on **Preferences**
3. Click on **Editor** tab
4. Click on **Choose...**
5. Select new editor (i.e. **notepad++.exe**
6. Click **Open** button

**Expected behavior**
After choosing the default editor would expect it to be available in right click of files, downloads, etc.  The previous preference of editor was eliminated during the upgrade.

 - OS: Windows 10
 - Version 8.4.1

**Additional context**
Problem appeared with the 8.4.1 version.

AliveDevil:
Fixes #13541

### DIFF
--- a/core/src/main/csharp/NativeMethods.txt
+++ b/core/src/main/csharp/NativeMethods.txt
@@ -25,6 +25,8 @@ IExtractIconW
 ILCreateFromPath
 IQueryAssociations
 IShellItem
+LoadLibrary
+LoadString
 MESSAGEBOX_RESULT
 PathParseIconLocation
 PBST_ERROR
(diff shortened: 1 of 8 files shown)


AliveDevil:
"Open With"-item is automatically named after system locale resource id 9016, is `Open With...`, and falls back to static string, if, for whatever obscure reason possible, there is no way of getting that resource string.

dkocher (approved):

PR #13567, opened 2022-07-13
Do not send any If header if pseudo locking is enabled.

ylangisc:
Do not send any If header if pseudo locking is enabled.

### DIFF
--- a/webdav/src/main/java/ch/cyberduck/core/dav/DAVCopyFeature.java
+++ b/webdav/src/main/java/ch/cyberduck/core/dav/DAVCopyFeature.java
@@ -25,6 +25,7 @@ import ch.cyberduck.core.features.Copy;
 import ch.cyberduck.core.features.Lock;
 import ch.cyberduck.core.http.HttpExceptionMappingService;
 import ch.cyberduck.core.io.StreamListener;
+import ch.cyberduck.core.preferences.HostPreferences;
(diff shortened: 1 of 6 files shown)


dkocher (commented):
The caller should nullify the `lockid` in the transfer status instead.

ylangisc:
> The caller should nullify the `lockid` in the transfer status instead.

What about making the check when looking up the feature at https://github.com/iterate-ch/cyberduck/blob/master/webdav/src/main/java/ch/cyberduck/core/dav/DAVSession.java#L342-L344?

dkocher:
> > The caller should nullify the `lockid` in the transfer status instead.
> 
> What about making the check when looking up the feature at https://github.com/iterate-ch/cyberduck/blob/master/webdav/src/main/java/ch/cyberduck/core/dav/DAVSession.java#L342-L344?

We can add a property like `webdav.lock.enable` which allows disabling lock feature ignoring support by the server.

dkocher:
> > > The caller should nullify the `lockid` in the transfer status instead.
> > 
> > 
> > What about making the check when looking up the feature at https://github.com/iterate-ch/cyberduck/blob/master/webdav/src/main/java/ch/cyberduck/core/dav/DAVSession.java#L342-L344?
> 
> We can add a property like `webdav.lock.enable` which allows disabling lock feature ignoring support by the server.

In #13580.

PR #13583, opened 2022-07-17
Review handling of 412 reply on upload

### LINKED ISSUE #13568: Avoid retries for DelayedHttpEntity
In https://github.com/iterate-ch/cyberduck/blob/master/webdav/src/main/java/ch/cyberduck/core/dav/DAVWriteFeature.java#L118-L130 we resend the initial `PUT` with a removed header. As we are completely stream based that leads to an empty stream causing zero bytes files uploaded for transfer with an unknown length (-1).

In case of known length we set a `Content-Length` header >0 but have an empty output stream causing the upload to hang.

dkocher:


ylangisc (changes requested):
* `Expect: 100-continue` would need to be enabled based on the lock settings. Currently statically disabled by default in `HttpConnectionPoolBuilder#createRequestConfig`
* enabling it doesn't resolve the issue as the first `PUT` request is sent with the `Expect: 100-continue` header but already with the full body. The `catch` clause is never reached as the stream is copied with the first `PUT` already.

dkocher:
> * `Expect: 100-continue` would need to be enabled based on the lock settings. Currently statically disabled by default in `HttpConnectionPoolBuilder#createRequestConfig`

`DAVWriteFeature` sets the `Expect` header based on the property `webdav.expect-continue` which is enabled by default.

ylangisc:
> > * `Expect: 100-continue` would need to be enabled based on the lock settings. Currently statically disabled by default in `HttpConnectionPoolBuilder#createRequestConfig`
> 
> `DAVWriteFeature` sets the `Expect` header based on the property `webdav.expect-continue` which is enabled by default.

Using a proxy you can see that the header field is removed by the http client as long as it is statically disabled in `HttpConnectionPoolBuilder#createRequestConfig`. As mentioned enabling it wouldn't resolve the issue either - the full body is sent in either way.

PR #13590, opened 2022-07-19
Load profiles only once when tab is selected.

dkocher:


ylangisc:
@AliveDevil Is there any change needed for Windows?

AliveDevil:
Profiles view in Cyberduck is already single-loaded (due to the Preferences window being singleton)
https://github.com/iterate-ch/cyberduck/blob/master/core/native/refresh/src/main/csharp/ViewModels/Preferences/Pages/ProfilesViewModel.cs#L51

On Mountain Duck the tab usually is loaded once it is activated, that is the tab is deselected, and reselected. With that flag it is only loaded once per lifetime of preferences-window.

So behavior, currently:
* On Cyberduck it is loaded once per App-lifetime
* On Mountain Duck everytime the preferences window is reopened and the profiles tab is selected, it gets refreshed

Should this be changed as well?

ylangisc (changes requested):
As the available profiles don't change often, I would suggest to change it to once per App-lifetime for all our ducks. Same for macOS.

AliveDevil:
Updated behavior in Windows implementation. Uses central registry/cache which lazy-loads, once requested.

AliveDevil:
Going to create PR to fix failing Windows builds.

AliveDevil:
Rebased to fix Windows-build.

ylangisc (changes requested):
As mentioned in https://github.com/iterate-ch/cyberduck/pull/13590#pullrequestreview-1050392593 the macOS part loading it once only per app lifetime is missing.

core/native/refresh/src/main/csharp/Services/ProfilesProvider.cs
```diff
@@ -0,0 +1,43 @@
...
+        private class InternalProfilesSynchronizeWorker : ProfilesSynchronizeWorker
+        {
+            private readonly TaskCompletionSource<IEnumerable<ProfileDescription>> completionSource;
+
+            public InternalProfilesSynchronizeWorker(ProtocolFactory protocolFactory, TaskCompletionSource<IEnumerable<ProfileDescription>> completionSource) : base(protocolFactory, ProfilesFinder.Visitor.Prefetch)
+            {
+                this.completionSource = completionSource;
+            }
+
+            public override void cleanup(object result) => completionSource.SetResult(Utils.ConvertFromJavaList<ProfileDescription>((java.util.Collection)result));
```
dkocher, line 40: On a network failure the `result` will be a set of profiles installed locally only. Thus it should not be cached.

dkocher:
> As mentioned in [#13590 (review)](https://github.com/iterate-ch/cyberduck/pull/13590#pullrequestreview-1050392593) the macOS part loading it once only per app lifetime is missing.

On a network failure the `result` of the `ProfilesSynchronizeWorker` will be a set of profiles installed locally only. Thus it should run once when opening a _Preferences_ window.

AliveDevil:
Then we should provide a separate cache for profiles (independent of client), and running ProfilesSynchronizeWorker everytime - doing a noop if one ProfilesSynchronizeWorker previously registered a successful remote listing.

dkocher:
> Then we should provide a separate cache for profiles (independent of client), and running ProfilesSynchronizeWorker everytime - doing a noop if one ProfilesSynchronizeWorker previously registered a successful remote listing.

I don't see much benefit of having a cache with a longer lifespan than the _Preferences_ window controller. It is fine to reload in this case.

ylangisc (approved):

PR #13609, opened 2022-07-27
Fix #13608.

### LINKED ISSUE #13608: Allow overwriting connection pool size per bookmark
Change from `Preferences` to `HostPreferences` for `SessionPool.create`[^1]. Allows limiting number of connections in FTP pool on Mountain Duck to prevent running out of allowed sessions on FTP servers.

[^1]: https://github.com/iterate-ch/cyberduck/blob/master/core/src/main/java/ch/cyberduck/core/SessionPoolFactory.java#L102

dkocher:


core/src/main/java/ch/cyberduck/core/SessionPoolFactory.java
```diff
@@ -97,9 +97,9 @@ public static SessionPool create(final ConnectionService connect, final Transcri
                     log.info(String.format("Create new pooled connection pool for %s", bookmark));
                 }
                 return new DefaultSessionPool(connect, trust, key, registry, transcript, bookmark)
-                        .withMinIdle(PreferencesFactory.get().getInteger("connection.pool.minidle"))
-                        .withMaxIdle(PreferencesFactory.get().getInteger("connection.pool.maxidle"))
-                        .withMaxTotal(PreferencesFactory.get().getInteger("connection.pool.maxtotal"));
+                        .withMinIdle(new HostPreferences(bookmark).getInteger("connection.pool.minidle"))
```
AliveDevil, line 100: Any reason to create three instances of HostPreferences? Can't we use one instance for all three properties?

AliveDevil (approved):

PR #13631, opened 2022-08-03
Review icon cache concurrency

AliveDevil:
Enumerating over a dictionary may result in `System.InvalidOperationException` when other operations modify the IconCache in parallel, shouldn't happen, but apparently it does.

This changes the behavior to lock-early (the cache), and force synchronized access to the cache. Trying to minimize over-writing already existing caches (`GetCacheExclusive` in tandem with `GetCache`).
Populating non-cached elements creates a write-lock over the whole import process. In regular usage this shouldn't block, as access to the cache usually happens on the main-thread (exceptions may apply).

The underlying Hashtable for registering typed objects isn't affected by this at all - this can still be written to in parallel (don't see an issue here, as it is always direct indexer-access, no enumeration).

### DIFF
--- a/core/native/refresh/src/main/csharp/Services/IconCache.cs
+++ b/core/native/refresh/src/main/csharp/Services/IconCache.cs
@@ -1,17 +1,19 @@
 ﻿using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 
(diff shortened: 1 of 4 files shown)


AliveDevil:
Mountain Duck ran just fine with this change. Assuming there is nothing wrong at the moment.

dkocher (approved):

PR #13668, opened 2022-08-17
Use same property for last update timestamp in millis.

dkocher:
Signed-off-by: David Kocher <dkocher@iterate.ch>


ylangisc (changes requested):
Does not compile on Windows.

ylangisc (approved):

PR #13686, opened 2022-08-23
Handle strange case that can occur when retrying requests in which th…

### LINKED ISSUE #13683: Interoperability with Netapp handling 501 reply for request with versioning parameter

Connect S3 from a Mac Book to a Netapp ONTAP 9.9.1.

**Describe the bug**
Everthing works fine. After an update to cyberduck 8.4.3. I got am error. Look at the screen shoot.

**To Reproduce**
If I reinstall Version 8.4.2. Everthing works fine.

**Expected behavior**

**Screenshots**
image

 - OS: 12.5.1

**Log Files**
Please include the [log file](https://docs.cyberduck.io/cyberduck/support/#logging-output). You can reach the logging output in _Preferences → Connection_.

**Additional context**

dkocher:
…e URI object has a null host value.

Signed-off-by: David Kocher <dkocher@iterate.ch>


ylangisc (changes requested):
We should find the reason here as it's probably reproducible, at least manually with Proxyman for example.

ylangisc (approved):

PR #13698, opened 2022-08-30
If there is an explicit key set in bookmark try to filter ssh agent i…

### LINKED ISSUE #13680: Prioritize identity set in bookmark before enumerating
There are servers disconnecting users after one failed authentication attempt.
Using SFTP Agent authentication can result in numerous failing attempts.

Is there a way where we can sort the list of identities found by priority (identity set in bookmark first, then all others)?

ylangisc:
…dentities based on the comment field which normally contains the key path. Resolves #13680.

### DIFF
--- a/ssh/src/main/java/ch/cyberduck/core/sftp/auth/SFTPAgentAuthentication.java
+++ b/ssh/src/main/java/ch/cyberduck/core/sftp/auth/SFTPAgentAuthentication.java
@@ -18,6 +18,7 @@ package ch.cyberduck.core.sftp.auth;
 import ch.cyberduck.core.AuthenticationProvider;
 import ch.cyberduck.core.DefaultIOExceptionMappingService;
 import ch.cyberduck.core.Host;
+import ch.cyberduck.core.Local;
 import ch.cyberduck.core.LoginCallback;
 import ch.cyberduck.core.exception.BackgroundException;
 import ch.cyberduck.core.sftp.SFTPExceptionMappingService;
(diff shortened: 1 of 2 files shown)


ssh/src/main/java/ch/cyberduck/core/sftp/auth/SFTPAgentAuthentication.java
```diff
@@ -74,4 +80,21 @@ public Boolean authenticate(final Host bookmark, final LoginCallback prompt, fin
...
+            final Collection<Identity> candidates = new ArrayList<>();
+            for(Identity identity : identities) {
+                if(identity.getComment() != null) {
+                    if(selected.getAbsolute().equals(new String(identity.getComment(), StandardCharsets.UTF_8))) {
+                        return Collections.singletonList(identity);
+                    }
+                }
+                candidates.add(identity);
+            }
+            return candidates;
```
dkocher, line 96: Same as returning original set `identities`?

ssh/src/main/java/ch/cyberduck/core/sftp/auth/SFTPAgentAuthentication.java
```diff
@@ -74,4 +80,21 @@ public Boolean authenticate(final Host bookmark, final LoginCallback prompt, fin
...
     }
+
+    protected Collection<Identity> filterIdentities(final Host bookmark, final Collection<Identity> identities) {
+        if(bookmark.getCredentials().isPublicKeyAuthentication()) {
+            final Local selected = bookmark.getCredentials().getIdentity();
+            final Collection<Identity> candidates = new ArrayList<>();
+            for(Identity identity : identities) {
+                if(identity.getComment() != null) {
+                    if(selected.getAbsolute().equals(new String(identity.getComment(), StandardCharsets.UTF_8))) {
+                        return Collections.singletonList(identity);
```
dkocher, line 91: Add log statement that only a single key from SSH agent is selected.

dkocher (changes requested):

dkocher (approved):

PR #13701, opened 2022-08-31
Localize lazily.

ylangisc:
Reduce CPU pressure for code paths that extensively use `NotFoundException`.


dkocher (changes requested):
We would want to change this for all subclasses of `BackgroundException`.

dkocher:
The `BundleLocale` implementation has a cache of `1000` entries for localized strings. There should be no performance penalty.

PR #13712, opened 2022-09-06
Save OAuth expiry in keychain instead of preferences

### LINKED ISSUE #13710: OAuth password flow does not save expiry for tokens

dkocher:


AliveDevil (commented):
Only on Linux/macOS (in case of CLI).
Windows already has persistent expiry tokens[^1] (both, in CLI and GUI).
[^1]: https://github.com/iterate-ch/cyberduck/pull/12918

AliveDevil (approved):

PR #13825, opened 2022-10-28
Return latest attributes after writing a file.

dkocher:


ylangisc (approved):

dkocher:
Skip because of possible transfer throughput regressions with additional attributes queried for every file after upload.

PR #13834, opened 2022-11-01
Bugfix/jni codepage

AliveDevil:


core/src/main/csharp/ch/cyberduck/core/proxy/SystemProxy.cs
```diff
@@ -46,9 +46,6 @@ public override ch.cyberduck.core.proxy.Proxy find(string host)
                 return ch.cyberduck.core.proxy.Proxy.DIRECT;
             }
 
-            // Hack to make Secur32/IWA work. With a non-us locale we get an invalid codepage (1) exception when using the native library.
-            Thread.CurrentThread.CurrentCulture = new CultureInfo("en-US");
```
AliveDevil, line 50: If this isn't fixed, I'm eating a broom.

ylangisc (approved):

PR #13849, opened 2022-11-04
Fix #13846.

### LINKED ISSUE #13846: Skip setting passphrase when sharing file is not intuitive
The [documentation](https://docs.cyberduck.io/cyberduck/share/) states

> Choose Cancel to create a public with no password protection

We should have a checkbox or additional button _Skip_ to create a share with no passphrase.

dkocher:
image
Screenshot 2022-11-14 at 13 57 43

### DIFF
--- a/windows/src/main/csharp/ch/cyberduck/ui/winforms/PasswordForm.cs
+++ b/windows/src/main/csharp/ch/cyberduck/ui/winforms/PasswordForm.cs
@@ -37,12 +37,17 @@ namespace Ch.Cyberduck.Ui.Winforms
                 Anchor = (((AnchorStyles.Left | AnchorStyles.Right))),
                 Text = "Save Password"
             };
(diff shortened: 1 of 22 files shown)


windows/src/main/csharp/ch/cyberduck/ui/winforms/PasswordForm.cs
```diff
@@ -37,14 +37,19 @@ public PasswordForm()
                 Anchor = (((AnchorStyles.Left | AnchorStyles.Right))),
                 Text = "Save Password"
             };
-            tableLayoutPanel.RowCount++;
```
AliveDevil, line 40: Wasn't required in my testing. Inserting into RowStyles is just fine.

AliveDevil (approved):

dkocher:
Updated documentation in iterate-ch/docs#291

PR #13860, opened 2022-11-08
Mark for implicit directory modification date change.

dkocher:


ftp/src/main/java/ch/cyberduck/core/ftp/FTPProtocol.java
```diff
@@ -36,6 +36,11 @@ public String getPrefix() {
         return String.format("%s.%s", FTPProtocol.class.getPackage().getName(), StringUtils.upperCase(this.getType().name()));
     }
 
+    @Override
+    public DirectoryTimestamp getDirectoryTimestamp() {
+        return DirectoryTimestamp.implicit;
```
AliveDevil, line 41: is it though?
I could see Microsoft FTP not adhering to this, or at least Microsoft FTP on NTFS drives, where modification dates are not propagated up the tree for directories.

dkocher: Yes, this also applies to UNIX. Modification dates are only updated for the parent folder and this assumption is not correct.

dkocher: Currently only Nextcloud/ownCloud have ETag and timestamps [propagate to the root](https://github.com/owncloud/client/wiki/Etags-and-file-ids#etag-change). Other WebDAV implementations *may* also support this.

AliveDevil: SFTP on Windows won't be implicit as well.

AliveDevil: So Windows would be direct parent folder only, not propagating up.

dkocher: Clarified in afff5eba that the definition only applies to direct parent folder.

ylangisc (approved):

PR #13870, opened 2022-11-14
Fix #13869.

### LINKED ISSUE #13869: Failure loading profile with duplicate custom property
From iterate-ch/profiles#18.

dkocher:


s3/src/test/resources/Wasabi (us-central-1).cyberduckprofile
```diff
@@ -46,6 +46,7 @@
         <key>Properties</key>
         <array>
             <string>s3.storage.class.options=STANDARD</string>
+            <string>s3.storage.class.options=STANDARD</string>
```
AliveDevil, line 49: Delete.

AliveDevil (changes requested):

s3/src/test/resources/Wasabi (us-central-1).cyberduckprofile
dkocher: Added to make sure test case fails without fix.

AliveDevil (approved):

PR #13917, opened 2022-11-30
Disable support because of missing support for custom attributes in P…

dkocher:
…ROPATCH.


ylangisc:
It's CTERA-65 not CTERA-61.

Issue #13723, opened 2022-09-13
Failure uploading file with diacritic in filename


jerry-ne:
**Describe the bug**
File with some czech letters is not uploaded.

**To Reproduce**
1. Open SFTP connection with UTF-8 encoding from Win10 (NTFS) to Ubuntu 20.04 LTS
2. Try to upload c:\temp\objížďka.mp4
3. I got Transfet incomplete and error msg: File is not found c:\temp\objížďka.mp4 Please contatct your file hosting suuport. /someting like that, because I have czech lang/
4. Click on dialog: More details:
1 OPENDIR
2 READDIR
3 READDIR
4 CLOSE
5 OPENDIR
6 READDIR
7 READDIR
8 CLOSE

**Expected behavior**
File starting upload.

**Log Files**
[cyberduck.log](https://github.com/iterate-ch/cyberduck/files/9560317/cyberduck.log)

dkocher:
```
2022-09-13 21:40:29,581 [Thread-0] DEBUG ch.cyberduck.core.unicode.AbstractUnicodeNormalizer - Normalized string C:\temp\objížďka.mp4 to C:\temp\objížďka.mp4
2022-09-13 21:40:35,656 [sFqARAhL-transfer-1] DEBUG ch.cyberduck.core.threading.DefaultFailureDiagnostics - Determine cause for failure BackgroundException{class=class ch.cyberduck.core.exception.LocalNotfoundException, file=null, message='Soubor nenalezen', detail='C:\temp\objížďka.mp4', cause='null'}
```

dkocher:
Should possibly set `local.normalize.unicode` [^1] to `false` on Windows.

[^1]: https://github.com/iterate-ch/cyberduck/blob/master/core/src/main/java/ch/cyberduck/core/Local.java#L96

PR #13951, opened 2022-12-06
Fix SystemLocal sanitization

### LINKED ISSUE #13933: Crash on Windows 11 Pro with IdentityAgent in ssh config
**Describe the bug**
Since version 8.3.0 in April, Cyberduck crashes on launch with my ssh config file. I originally thought my problem was #13672, but after realizing the fix for that was supposed to have been released already, I actually checked my crash reports and narrowed this down today. As of 8.4.5.38423 and 8.5.1.38745, the problem goes away if I comment out my IdentityAgent line. Fortunately, I don't think I actually need it - I don't use the SSH agent, so the following ProxyCommand line, which doesn't cause me any problems, is the important one.

**To Reproduce**
1. Have an ssh config with a line that starts with `IdentityAgent //` 
2. Try to launch Cyberduck

If the next character of the IdentityAgent is a period, the program crashes. If it is anything else, it seems to hang indefinitely with a process running but the UI never appears.

The reason I had the configuration in question was because Krypton Authenticator adds the following (path redacted):

```
# Added by Krypton
Host *
	IdentityAgent \\.\pipe\krd-agent
	ProxyCommand path\to\krssh.exe %h %p
```

but I had a _different_ tool - unfortunately I've forgotten what it was; it may have been a Visual Studio Code plugin - which worked if I changed the backslashes in the IdentityAgent line to forward slashes but not with the configuration as above.

Since I have a workaround, I've submitted this mostly in case it's a simple parsing issue that (A) is easy to fix and (2) might trip people up in a less obscure case. I may even come back later and try to whip up a pull request. My assumption is that the problem is in handling that kind of string in `LocalFactory.get()`, but I ran out of time to investigate this today before I figured out where LocalFactory lives in the codebase.

**Expected behavior**
Cyberduck working; if necessary, a convenient and documented way to tell Cyberduck not to try to parse my ssh config or to point it at a different one (which might already exist and I failed to find it...).

 - OS: Windows 11 Pro and Windows 11 Pro Insider Preview
 - Version: all attempts from at least 10.0.22593 through 10.0.25247

**Log Files**
[638055899217769141.txt](https://github.com/iterate-ch/cyberduck/files/10143983/638055899217769141.txt)

AliveDevil:


core/src/main/csharp/ch/cyberduck/core/local/SystemLocal.cs
```diff
@@ -83,62 +82,119 @@ public override bool isSymbolicLink()
...
+                    char c = (char)reader.Read();
+                    if (c == Path.AltDirectorySeparatorChar || c == Path.DirectorySeparatorChar)
+                    {
+                        writer.Write(segmentBuilder.ToString().Trim());
+                        writer.Write(Path.DirectorySeparatorChar);
+                        segmentBuilder.Clear();
+                    }
+                    else if (Array.IndexOf(invalid, c) != -1)
+                    {
+                        segmentBuilder.Append('_');
```
AliveDevil, line 186: @ylangisc Still UrlEncode invalid characters? Asof Windows 10 1803 '%'[^ms-docs] is a valid filename character. Thus rendering this invalid on any system which is not Windows 10 1803 (or newer, which is true for this[^commit] as well)
[^ms-docs]: https://learn.microsoft.com/troubleshoot/windows-client/networking/url-encoding-unc-paths-not-url-decoded
[^commit]: b169fcfbbbf2c3a9a2a9ca239dbe34d6db09347c

dkocher:
Does this relate to #13723 as well?

AliveDevil:
Fixes WSL drag/drop.
Next: UNC

AliveDevil:
For Dragging onto UNC we'd need to implement IStorage[^IStorage], providing an enumeration of all Elements to be transferred.
[^IStorage]: https://learn.microsoft.com/en-us/windows/win32/api/objidl/nn-objidl-istorage

This must hook between Transfer#prepare (listing all files) and actually downloading files. Could we potentially just ignore the Transfer-window and directly drop into Background-processing for this scenario?

dkocher:
> This must hook between Transfer#prepare (listing all files) and actually downloading files. Could we potentially just ignore the Transfer-window and directly drop into Background-processing for this scenario?

Could possibly be implemented as a `Bulk` [^1] feature.


[^1]: https://github.com/iterate-ch/cyberduck/blob/master/core/src/main/java/ch/cyberduck/core/features/Bulk.java

AliveDevil:
Fix for #5330 is scheduled for later fix.

core/src/main/java/ch/cyberduck/core/transfer/download/AbstractDownloadFilter.java
```diff
@@ -198,9 +196,9 @@ public TransferStatus prepare(final Path file, final Local local, final Transfer
                     // Free space on disk
                     long space = 0L;
                     try {
-                        space = Files.getFileStore(Paths.get(local.getParent().getAbsolute())).getUsableSpace();
+                        space = Paths.get(local.getParent().getAbsolute()).toFile().getUsableSpace();
```
dkocher, line 199: Why is this change required?

dkocher (changes requested):

core/src/main/java/ch/cyberduck/core/transfer/download/AbstractDownloadFilter.java
AliveDevil:
```
Files.getFileStore("\\…") == null
```
With `Paths.get().toFile()` we get access to the FileStore of that file, thus can see usablespace for remotes.

core/src/test/csharp/ch/cyberduck/core/local/SystemLocalTest.cs
```diff
@@ -0,0 +1,43 @@
...
+        }
+
+        [Test]
+        public void TestWslPath()
+        {
+            CoreLocal local = new SystemLocal(WSL_PATH);
+            Assert.AreEqual(WSL_PATH, local.getAbsolute());
+        }
+    }
+}
```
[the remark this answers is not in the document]
AliveDevil: Clarify `Path with disk label`.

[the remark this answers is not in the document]
dkocher: `C:\absolute\path\to\file`

core/src/main/java/ch/cyberduck/core/transfer/download/AbstractDownloadFilter.java
dkocher: Do not understand your comment. It is about determining free disk space on the target volume.

AliveDevil: Yes, correct. If downloading to a network drive/unc path in the future `Files.getFileStore` will result in `null` being returned.

That is: This will fail to determine disk space and disable segmented downloads to `\\wsl$`, `\\wsl.localhost`, `\\?\C:\Long\Path`.

dkocher (approved):

dkocher (approved):

PR #13960, opened 2022-12-08
Special handling for username obtained from active directory stripping domain for username in default path

dkocher:


nextcloud/src/main/java/ch/cyberduck/core/nextcloud/NextcloudHomeFeature.java
```diff
@@ -37,19 +35,20 @@ public NextcloudHomeFeature(final Host bookmark) {
...
+        String username = bookmark.getCredentials().getUsername();
+        if(StringUtils.isBlank(username)) {
             if(log.isWarnEnabled()) {
                 log.warn(String.format("Missing username for %s", bookmark));
             }
             return null;
         }
         final Path workdir = new Path(new Path(String.format("/remote.php/dav/%s", files.name()), EnumSet.of(Path.Type.directory)),
-                URIEncoder.encode(bookmark.getCredentials().getUsername()), EnumSet.of(Path.Type.directory));
+                username, EnumSet.of(Path.Type.directory));
```
ylangisc, line 51: No need to URI encode `username`?

dkocher: This is handled separately when doing requests but file paths in the model are never encoded.

ylangisc (approved):

PR #13968, opened 2022-12-11
Do not persist setting to connect without virtual host style requests when determined because of DNS failure.

### LINKED ISSUE #13528: Cannot read container configuration
Trying to open a bucket I get this error.  This bug is not present in previous version 8.3.0, I had to downgrade the version to resolve this issue.  Please take note.
My bucket properties hasn't been changed in years; so I gues this is a cyberduck problem.  Thank you. 

Captura de pantalla 2022-06-29 181618

dkocher:


ylangisc (changes requested):
Add test for redirect.

PR #13983, opened 2022-12-14
Service on protocol level to use when comparing files & folders

dkocher:


core/src/main/java/ch/cyberduck/core/synchronization/SizeComparisonService.java
```diff
@@ -27,24 +29,27 @@ public class SizeComparisonService implements ComparisonService {
...
+            if(remote.getSize() == 0 && local.getSize() == 0) {
+                return Comparison.equal;
+            }
+            if(remote.getSize() == 0) {
+                return Comparison.local;
+            }
+            if(local.getSize() == 0) {
+                return Comparison.remote;
+            }
+            if(remote.getSize() == local.getSize()) {
```
ylangisc, line 47: Move up as first check to eliminate the 0 bytes check.

ylangisc (approved):

PR #13997, opened 2022-12-15
Rewrite object when attempting to set timestamp earlier than already …

dkocher:
…set.


ylangisc:
Did some performance testing with rewriting larger objects. 200MB takes ~250ms.

ylangisc (approved):

PR #14016, opened 2022-12-22
Fix NPE for broken bookmark files and make sure that the stream is no…

ylangisc:
…t reused.


dkocher:
Also review `DictionaryLicense` please.

dkocher (approved):

PR #14029, opened 2023-01-02
Fixes regression in #13951

### LINKED ISSUE #13933: Crash on Windows 11 Pro with IdentityAgent in ssh config

AliveDevil:
https://github.com/iterate-ch/cyberduck/pull/13951

### DIFF
--- a/core/src/main/csharp/ch/cyberduck/core/local/SystemLocal.cs
+++ b/core/src/main/csharp/ch/cyberduck/core/local/SystemLocal.cs
@@ -88,16 +88,25 @@ namespace Ch.Cyberduck.Core.Local
             using StringWriter writer = new();
             using StringReader reader = new(name);
 
-            if (reader.Peek() is int p && (p == Path.DirectorySeparatorChar || p == Path.AltDirectorySeparatorChar))
+            switch (reader.Peek())
             {
-                if (!ReadUnc(reader, writer))
-                {
-                    return null;
-                }
-            }
-            else if (!ReadDriveLetter(reader, writer, false))
-            {
-                return null;
+                case int p when p == Path.DirectorySeparatorChar || p == Path.AltDirectorySeparatorChar:
+                    if (!ReadUnc(reader, writer))
+                    {
+                        return null;
+                    }
+                    break;
+
+                case int p when p == '~':
+                    writer.Write((char)reader.Read());
+                    break;
+
+                default:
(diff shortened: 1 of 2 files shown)


dkocher (changes requested):
Please add previously failing test.

AliveDevil:
Is there: [40676ba](https://github.com/iterate-ch/cyberduck/pull/14029/commits/40676ba0ed2b43444c99ab783a0ea10104a74eb9)

dkocher (approved):

PR #14046, opened 2023-01-05
Fix #14045.

### LINKED ISSUE #14045: Missing AWS S3 Bucket Regions
Add the missing bucket regions:
- Asia Pacific (Hyderabad): ap-south-2
- Asia Pacific (Jakarta): ap-southeast-3
- China (Beijing): cn-north-1
- China (Ningxia): cn-northwest-1
- Europe (Spain): eu-south-2
- Europe (Zurich): eu-central-2
- Middle East (UAE): me-central-1

dkocher:


s3/src/main/java/ch/cyberduck/core/s3/S3Protocol.java
```diff
@@ -101,17 +101,22 @@ public Set<Location.Name> getRegions() {
...
             new S3LocationFeature.S3Region("ap-northeast-2"),
             new S3LocationFeature.S3Region("ap-northeast-3"),
             new S3LocationFeature.S3Region("ap-south-1"),
+            new S3LocationFeature.S3Region("ap-south-2"),
             new S3LocationFeature.S3Region("ap-east-1"),
             new S3LocationFeature.S3Region("sa-east-1"),
             new S3LocationFeature.S3Region("me-south-1"),
-            new S3LocationFeature.S3Region("af-south-1")
+            new S3LocationFeature.S3Region("af-south-1"),
+            new S3LocationFeature.S3Region("me-central-1")
```
AliveDevil, line 119: me-south, then af-south, then me-central?
Can we alphabetically sort this list?

dkocher: Sorting is expected to be done in the controller preparing the regions for display.

AliveDevil: That is OK, the issue I have here is that every other region is grouped together (eu, ap, us), but `me` is split with `af` in between.

This list should really be sorted for easier human-editability. Doing a comparison-search by eye is more difficult than doing an alphabetical search. Thus there is a chance, that if this is not ordered duplicate or missed entries aren't caught.

AliveDevil (approved):

PR #14153, opened 2023-01-30
Fix file checkout in OneDrive/Sharepoint

AliveDevil:


onedrive/src/test/java/ch/cyberduck/core/onedrive/GraphLockFeatureTest.java
```diff
@@ -65,4 +66,30 @@ public void testLock() throws Exception {
         }
         new GraphDeleteFeature(session, fileid).delete(Collections.singletonList(file), new DisabledLoginCallback(), new Delete.DisabledCallback());
     }
+
+    @Test
+    @Ignore("Same user can checkout file multiple times (╯‵□′)╯︵┻━┻")
```
dkocher, line 71: Remove ignore flag with fix?

AliveDevil: That would require, that we can lock a file with two different users in unit tests.
This is for documentation purposes only, as there is no error when the same user locks a file again.

AliveDevil: This[^1] won't throw - at all - for the same user, but throws if this is used by different users.

[^1]: https://github.com/iterate-ch/cyberduck/pull/14153/files#diff-38904e83dde33bc3d8f22c607b4543567f639f9e9bb1133bd52ccda67e810305R86

dkocher (approved):

PR #14178, opened 2023-02-03
Fix #14084.

### LINKED ISSUE #14084: Moving / copying file between windows belonging to same server+user should perform entirely server-side filesystem operation
## Reproduction

1. Open a new window and connect to a SFTP host.
2. You get a listing, e.g.: `dir-a, dir-b, dir-c.`
3. Right click `dir-a` → "New browser".
   - This opens a new browser window (instance) on the SAME server as the SAME user.
4. In your first open window, navigate into `dir-b`.
5. Copy `file1.mp4` from `dir-a` to `dir-b`.

## Expected: This happens in an instant.
- Cyberduck realizes this is the SAME filesystem on the SAME server with the SAME user and hence just triggers server-side filesystem move operation move file1.mp4 from dir-a to dir-b.
- Move operation: Even if a multi gigabyte file this happens in an instance (as just the filesystem TOC gets rewritten).
- Copy operation:
  - Server uses Copy on write (CoW) filesystem: This also happens in an instant.
  - Server does not use a Copy on write (CoW) filesystem: Happens with local HDD/SSD I/O speed, which almost always will be faster than any duplex network transfer (download to client and re-upload to server).
## Actually: This is painfully slow
- As Cyberduck performs a download of `file1.mp4` from `dir-a` to your computer's `/tmp/` directory and then re-uploads `file1.mp4` to `dir-b` on the server.

## Best Practises
- This is technically possible and available in other FTP software.
- I tried the Transmit trial version and there you can easily copy between windows which belong to the same server in that fashion (file move operations run entirely server side).

## Workaround
- Use only a single Cyberduck window.
- Use the folder-tree to expand `dir-a` and `dir-b` and then drag-n-drop `file1.mp4` from `dir-a` to `dir-b`.
- This works, but can be quite cumbersome in setups with long directory/file-lists.
- Lots of scrolling up and down… you gather many mouse kilometers…

dkocher:


osx/src/main/java/ch/cyberduck/ui/cocoa/datasource/BrowserTableDataSource.java
```diff
@@ -418,19 +419,21 @@ public boolean acceptDrop(final NSTableView view, final Path destination, final
...
-                if(pasteboard.getBookmark().compareTo(controller.getSession().getHost()) != 0) {
-                    // Drag to browser windows with different session or explicit copy requested by user.
-                    final Host target = controller.getSession().getHost();
-                    controller.transfer(new CopyTransfer(pasteboard.getBookmark(), target, files),
-                        new ArrayList<Path>(files.values()), false);
-                }
-                else if(info.draggingSourceOperationMask().intValue() == NSDraggingInfo.NSDragOperationCopy.intValue()) {
-                    // The file should be copied
-                    new CopyController(controller).copy(files);
+                if(new AbstractHostCollection.ProfilePredicate(pasteboard.getBookmark()).test(controller.getSession().getHost())) {
```
ylangisc, line 422: Any need to include the default path in the test as `ProfilePredicate` does not contain it?

dkocher: The fix is to *not* include the default path as this is set when choosing _New Browser_ in the context menu.

ylangisc (approved):

PR #14184, opened 2023-02-03
Make search with glob pattern case-insensitive.

### LINKED ISSUE #14179: Glob pattern operators asterisk * and question mark ? do not work
## Facts
- #13781 was added in [8.4.5](https://github.com/iterate-ch/cyberduck/milestone/204)
- The [documentation](https://docs.cyberduck.io/cyberduck/browser/#filter-and-search) says:
> Search is case-insensitive and does look for matching sequences in the filename supporting glob patterns * and ?.

## Environment
- Client: Cyberduck 8.5.5 (39213) on macOS 11.7.2 Big Sur
- Server: SFTP, on custom port (5-digits, assigned by my webhost), authentication with SSH key

## Bug
- Both glob pattern operators `*` and `?` do obviously not work, see screen recording (1 image every 3 secs):

https://user-images.githubusercontent.com/737143/216614723-71b2cb79-78c9-4cfe-a3bf-584e0211560c.mp4

dkocher:


setup/app/Info.plist
```diff
@@ -275,7 +275,7 @@
             <key>WorkingDirectory</key>
             <string>$APP_PACKAGE/Contents/Resources</string>
             <key>VMOptions</key>
-            <string>$JVM_RUNTIME_ARGS</string>
+            <string>-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=*:5005 $JVM_RUNTIME_ARGS</string>
```
ylangisc, line 278: Unwanted change.

core/src/main/java/ch/cyberduck/ui/browser/SearchFilter.java
```diff
@@ -72,8 +72,9 @@ public String toString() {
     public Pattern toPattern() {
         final StringBuilder pattern = new StringBuilder();
         pattern.append("\\A");
```
ylangisc, line 74: Leading the regex with `(?i)` does the trick. No other lower case operations needed.

ylangisc (changes requested):

core/src/main/java/ch/cyberduck/ui/browser/SearchFilter.java
dkocher: Added to iterate-ch/docs#320.

PR #14204, opened 2023-02-07
CBC based ciphers are considered to be weak and thus not enabled by d…

ylangisc:
…efault in OpenSSH using OpenSSL 3.

### DIFF
--- a/ssh/src/main/java/ch/cyberduck/core/sftp/SFTPSession.java
+++ b/ssh/src/main/java/ch/cyberduck/core/sftp/SFTPSession.java
@@ -19,6 +19,7 @@ package ch.cyberduck.core.sftp;
  */
 
 import ch.cyberduck.core.*;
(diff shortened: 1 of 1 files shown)


ylangisc:
https://github.com/proftpd/proftpd/issues/1595

ssh/src/main/java/ch/cyberduck/core/sftp/SFTPSession.java
```diff
@@ -433,6 +437,21 @@ public void disconnect() {
         super.disconnect();
     }
 
+    private List<Named<Cipher>> prioritizeCTRoverCBC(final List<Named<Cipher>> factories) {
+        final List<Named<Cipher>> reordered = new ArrayList<>();
```
dkocher, line 441: Sort existing list instead using comparator.

dkocher (changes requested):

dkocher (approved):

PR #14205, opened 2023-02-07
Make sure an agent is running before adding it as authentication method.

### LINKED ISSUE #14196: On Windows IdentitiesOnly with public identity file through SSH agent not authenticating
SFTP access.
It works from the Windows shell, but I get an error with Cyberduck:

```
Host hhh
  HostName hhh.cloudapp.azure.com
  User hhh-admin
  IdentityFile \Users\hhh\.ssh\hhh-admin-key.pub
  IdentitiesOnly yes
```

```
ch.cyberduck.core.AbstractExceptionMappingService - No message for failure net.schmizz.sshj.common.SSHException
ch.cyberduck.core.threading.DefaultFailureDiagnostics - Determine cause for failure BackgroundException{class=class ch.cyberduck.core.exception.BackgroundException, file=null, message='Accesso fallito.', detail='Sconosciuto.', cause='net.schmizz.sshj.common.SSHException'}
```

ylangisc:
Fixes #14196.

### DIFF
--- a/ssh/src/main/java/ch/cyberduck/core/sftp/openssh/OpenSSHAgentAuthenticator.java
+++ b/ssh/src/main/java/ch/cyberduck/core/sftp/openssh/OpenSSHAgentAuthenticator.java
@@ -55,7 +55,11 @@ public class OpenSSHAgentAuthenticator extends AgentAuthenticator {
     @Override
     public Collection<Identity> getIdentities() {
         if(null == proxy) {
-            log.warn(String.format("SSH agent %s is not running", this));
+            log.warn("Missing proxy reference");
+            return Collections.emptyList();
+        }
+        if(!proxy.isRunning()) {
+            log.warn(String.format("Agent %s not running", this));
             return Collections.emptyList();
         }
         if(log.isDebugEnabled()) {
--- a/ssh/src/main/java/ch/cyberduck/core/sftp/auth/SFTPAgentAuthentication.java
+++ b/ssh/src/main/java/ch/cyberduck/core/sftp/auth/SFTPAgentAuthentication.java
@@ -71,6 +71,10 @@ public class SFTPAgentAuthentication implements AuthenticationProvider<Boolean> 
         if(log.isDebugEnabled()) {
             log.debug(String.format("Login using agent %s for %s", agent, bookmark));
         }
+        if(agent.getIdentities().isEmpty()) {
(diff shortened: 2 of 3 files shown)


ssh/src/main/java/ch/cyberduck/core/sftp/SFTPSession.java
```diff
@@ -278,19 +278,26 @@ private void authenticate(final SSHClient client, final Host host, final LoginCa
             final String identityAgent = new OpenSSHIdentityAgentConfigurator().getIdentityAgent(host.getHostname());
             switch(Factory.Platform.getDefault()) {
                 case windows:
```
dkocher, line 280: These checks are redundant it seems if proper checks for agent to be available in implementation for `getIdentities`

ssh/src/main/java/ch/cyberduck/core/sftp/auth/AgentAuthenticator.java
```diff
@@ -25,4 +25,9 @@ public abstract class AgentAuthenticator {
     public abstract Collection<Identity> getIdentities();
 
     public abstract AgentProxy getProxy();
+
+    public boolean isAvailable() {
```
dkocher, line 29: Move this check instead to `getIdentities` of implementations.

dkocher (changes requested):

dkocher (approved):

ssh/src/main/java/ch/cyberduck/core/sftp/openssh/OpenSSHAgentAuthenticator.java
```diff
@@ -55,7 +55,11 @@ public AgentProxy getProxy() {
     @Override
     public Collection<Identity> getIdentities() {
         if(null == proxy) {
-            log.warn(String.format("SSH agent %s is not running", this));
+            log.warn("Missing proxy reference");
+            return Collections.emptyList();
+        }
+        if(!proxy.isRunning()) {
```
blaineam, line 61: I think this additional conditional is breaking 1password for non IdentityFile uses. I don't think this check is needed for the original issue either.

blaineam (commented):
This PR broke Mountain Duck for my 1Password 8 SSH setup. I'm pretty sure the section that broke it is the commented block.

PR #14226, opened 2023-02-13
Files.com API is case-insensitive

dkocher:


brick/src/test/java/ch/cyberduck/core/brick/BrickTouchFeatureTest.java
```diff
@@ -28,25 +27,21 @@
...
 import java.util.EnumSet;
 
-import static org.junit.Assert.fail;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 public class BrickTouchFeatureTest extends AbstractBrickTest {
 
-    @Test(expected = ConflictException.class)
+    @Test
     public void testCaseSensitivity() throws Exception {
```
ylangisc, line 35: Test should assert that `list.size() == 1` to ensure that only one file exists.

ylangisc (approved):

Issue #12917, opened 2022-02-24
IPv6 not working for hosts with both A and AAAA entries


ylangisc:
According to the [documentation](https://docs.cyberduck.io/cyberduck/connection/?highlight=prefer%20ipv6#prefer-ipv6-adresses-of-dns-lookups) you can prioritize IPv6 over IPv4 for hosts that have both A and AAAA DNS entries. This setting in turn sets the JAVA system property `java.net.preferIPv6Addresses` to `true`.

`InetAddress.getByName(hostname)` does return an IPv6 address for hosts with an A entry only. For hosts with both entries the IPv4 address is returned. The system property above does not have any effect in that case. Also refer to https://stackoverflow.com/a/11974740 which proposes to use `InetAddress.getAllByName(hostname)` or a 3rd party library.

Affected code in [Resolver](https://github.com/iterate-ch/cyberduck/blob/master/core/src/main/java/ch/cyberduck/core/Resolver.java#L62) class.

PR #14285, opened 2023-02-26
Preferences.java: Set preferIPv6Addresses to default true

### LINKED ISSUE #14286: Mountain Duck: no attempt to use IPv6 addresses
**Describe the bug**
When a Mountain Duck client runs on a dual stack network, connecting to a dual stack server, IPv4 is used instead of IPv6.

**To Reproduce**
1. Be on a dual stack network.
2. Access a dual stack WebDAV server.
3. Logs show connections over IPv4.

**Expected behavior**
IPv6 preferred.

 - OS: Windows
 - Version Latest

**Additional context**
Our IPv6 path is faster than IPv4, we have some disappointed users.

jpds:
IPv6 is now the majority IP version in various parts of the world:

- https://www.google.com/intl/en/ipv6/statistics.html
- https://stats.labs.apnic.net/ipv6

As such, this PR sets the app behavior to prefer IPv6 connections by default.


core/src/main/java/ch/cyberduck/core/http/HttpConnectionPoolBuilder.java
```diff
@@ -180,7 +180,6 @@ public HttpClientBuilder build(final Proxy proxy, final TranscriptListener liste
                 new BackportWindowsNegotiateSchemeFactory(null) :
                 new SPNegoSchemeFactory())
             .register(AuthSchemes.KERBEROS, new KerberosSchemeFactory()).build());
-        configuration.setDnsResolver(new CustomDnsResolver());
```
dkocher, line 183: What is the intention with removing the custom resolver implementation?

dkocher:
This setup was introduced with #12917.

core/src/main/java/ch/cyberduck/core/Resolver.java
```diff
@@ -46,16 +46,6 @@ public final class Resolver {
     private final ThreadFactory threadFactory
             = new NamedThreadFactory("resolver");
 
-    private final boolean preferIPv6;
-
-    public Resolver() {
-        this(PreferencesFactory.get().getBoolean("connection.dns.ipv6"));
-    }
-
-    public Resolver(final boolean preferIPv6) {
```
dkocher, line 55: I would expect test failures in `ResolverTest`

dkocher (changes requested):

defaults/src/main/resources/default.properties
```diff
@@ -547,8 +547,6 @@ transfer.connection.retry.delay=0
 connection.hostname.default=
 # Convert hostname to Punycode
 connection.hostname.idn=true
-# java.net.preferIPv6Addresses
-connection.dns.ipv6=false
```
dkocher, line 551: We should probably only just change this default to `true`.

core/src/main/java/ch/cyberduck/core/http/HttpConnectionPoolBuilder.java
jpds: Removing unnecessary code? This shouldn't be necessary anymore as the operating system should be the one making the decision of which IP version is used.

defaults/src/main/resources/default.properties
jpds: I'd really rather that the OS handles this (instead of implementing application-level hacks that shouldn't be there in the first place).

core/src/main/java/ch/cyberduck/core/http/HttpConnectionPoolBuilder.java
jpds: For context, this came from #13399 which specifically was added for IPv6 preference.

dkocher: I am hesitant as this will cause us to use two different resolver implementations in our [generic connection setup](https://github.com/iterate-ch/cyberduck/blob/e255b0c424d6e24a528fafdbe2acc525c48c8b8c/core/src/main/java/ch/cyberduck/core/LoginConnectionService.java#L39) and for the HTTP connection pool. In the best case this will only (but still) result in different error messages missing localization.

jpds: I don't understand (and obviously I'm not familiar with the entire codebase myself):

- what I see is that you have a Resolver in `LoginConnectionService.java` which _doesn't_ use a custom DNS resolver for IPv6 preference
- inside `HttpConnectionPoolBuilder.java` - I've reverted the custom DNS resolver

So instead of causing there to be two different resolver implementations - they're actually now the same? I also don't see any reference to a resolver within the `http` directory.

defaults/src/main/resources/default.properties
dkocher: I would prefer too, but this also depends on defaults in the Java Runtime.

core/src/main/java/ch/cyberduck/core/Resolver.java
```diff
@@ -73,9 +63,6 @@ public void run() {
                 try {
                     final InetAddress[] allByName = InetAddress.getAllByName(hostname);
                     Arrays.stream(allByName).findFirst().ifPresent(resolved::set);
-                    if(preferIPv6) {
```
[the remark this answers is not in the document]
jpds: it shouldn't - because this is about preferring IPv6 addresses over IPv4 - not forcing connections exclusively to IPv6.

The code already uses the `getAllByName()`  that @ylangisc has referenced in the issue report.

defaults/src/main/resources/default.properties
jpds: Even Sun Java supported this back in 2002: https://docs.oracle.com/javase/7/docs/technotes/guides/net/ipv6_guide/index.html - if there's a JRE out there that has no IPv6 support two decades later - it can only be described as: objectively garbage.

We should merge this, get a beta build out that users can heavily test.

dkocher: Sure the IPv6 stack is available but we are just talking about order preferences in name resolution always prioritizing IPv4.

> By default, we would prefer IPv4 addresses over IPv6 addresses, i.e., when querying the name service [^1] 

[^1]: https://docs.oracle.com/javase/7/docs/technotes/guides/net/ipv6_guide/index.html#ipv6-related

core/src/main/java/ch/cyberduck/core/Resolver.java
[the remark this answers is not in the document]
dkocher: This is correct, given we would set the default of `java.net.preferIPv6Addresses` to `true`, no additional sorting should be required for IPv6 addresses to be preferred.

AliveDevil:
Need to check what behavior .NET WebRequest and TcpClient default to (on .NET 4.7.1)[^1]

Otherwise, if Java-core defaults to IPv6, and our TcpReachability tests for IPv4, this may result in issues.

[^1]: https://g.iterate.ch/projects/ITERATE/repos/cyberduck/browse/core/src/main/csharp/ch/cyberduck/core/diagnostics/TcpReachability.cs

defaults/src/main/resources/default.properties
dkocher: Relates to #16723.

PR #14319, opened 2023-03-07
Fill byte array with null values until buffer length in case we read …

ylangisc:
…over file length.

### DIFF
--- a/core/src/main/java/ch/cyberduck/core/io/FileBuffer.java
+++ b/core/src/main/java/ch/cyberduck/core/io/FileBuffer.java
@@ -24,6 +24,8 @@ import ch.cyberduck.core.local.TemporaryFileServiceFactory;
 
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.io.input.NullInputStream;
+import org.apache.commons.lang3.concurrent.ConcurrentException;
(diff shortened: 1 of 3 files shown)


core/src/main/java/ch/cyberduck/core/io/FileBuffer.java
```diff
@@ -62,7 +62,16 @@ public synchronized int read(final byte[] chunk, final Long offset) throws IOExc
...
-                return file.read(chunk, 0, (int) (file.length() - offset));
+                final int read = file.read(chunk, 0, (int) (file.length() - offset));
+                if(read + offset < file.length()) {
+                    // Less read than we have in file - honour read interface
+                    return read;
+                }
+                //
+                final int missing = chunk.length - read;
+                final int available = (int) Math.min(missing, this.length - file.length());
+                System.arraycopy(new byte[available], 0, chunk, read, missing);
```
dkocher, line 73: Last parameter should be `available`?

dkocher (changes requested):

ylangisc:
Reuse of buffer after closing it is no longer possible (`write` after `close`). Refer to `ec67f66f9f7e6c0a9ecfc58e9ae22a5db46b96ba` in Mountain Duck.

dkocher (approved):

PR #14431, opened 2023-03-27
Hide directory placeholders only containing hidden files

dkocher:


ylangisc (approved):

Cherry:
This PR appears to have caused a large performance regression when listing large B2 directories and files. I'm confirming and creating an issue now.

metadaddy:
I'm not sure that it's possible to successfully implement this functionality. Consider a bucket with millions of objects with keys of the form `folder00000001/foo`, `folder00000002/foo`, `folder00000003/foo`, etc. This is not a particularly contrived example - analytical data can have this kind of structure. Every 'folder name' will need to be inspected, via a call to `b2_get_file_info`, to determine whether it should be shown.

PR #14543, opened 2023-04-17
No more Fody

AliveDevil:
Everything can be represented in C#, thus don't use Fody.

Don't need magic-property generation.

Changes:
- Embed Bonjour COM interop types
- Remove InlineIL with simpler constructs (new UnscopedRef-feature in CSC/Roslyn)
- Remove automatic reactive property generation
- Format some CsWin32-generated files


dkocher:
@AliveDevil Squashing commits?

dkocher (approved):

PR #14568, opened 2023-04-23
Support AWS S3 sso_session configuration

### LINKED ISSUE #14276: AWS SSO Profile config has new format
Regarding the AWS CLI SSO login and the hints how to use it in #12890

The AWS cli now supports an SSO Session name which puts some config values in its own block and changes the calculated hash.
See https://github.com/boto/botocore/blob/1.29.77/botocore/credentials.py#L2078

To get the old behavior when configuring the SSO session name needs to be empty.
```aws configure sso
SSO session name (Recommended):
WARNING: Configuring using legacy format (e.g. without an SSO session).
Consider re-running "configure sso" command and providing a session name.```

Could you add the new hash format too?

yuzawa-san:
mirrored the boto cache key code (link in code comment) to use sso_session instead of sso_start_url. this is backwards compatible since it is conditional based on what is the config properties.

fixes https://github.com/iterate-ch/cyberduck/issues/14276


dkocher:
Could you rebase to `master`, I had to [revert](https://github.com/iterate-ch/cyberduck/commit/313b56d64861a4e49ac5abec8fcf1a940ab58920) publishing tests results which fail for pull requests with `Resource not accessible by integration`.

dkocher (approved):

PR #14571, opened 2023-04-24
Use own preferences implementation

### LINKED ISSUE #14566: S3 Default ACL Settings reset on update of Windows Store version
**Describe the bug**
When the Windows Store version updates, it resets the Preferences-> S3 -> Default ACL option back to a default value of "private". I need this set to public-read for my uploads or else files are inaccessible by my users, so this is a repeating source of frustration for me.

**To Reproduce**
1. Go to Preferences
2. Click on S3
3. Scroll down to Default ACL
4. Set it to public-read
5. Wait for cyberduck to get updated.
6. Go back and see the setting was reset.

**Expected behavior**
Retain my previous settings from before the update.

 - OS: Windows 11 22H2
 - Version 8.5.9

AliveDevil:
- Migrate existing config over to `%AppData%\DataFolder\ProductName.user.config`
- Setup JNA on module-init (removes need for calling `SetJnaPath()`)
- DataFolder requires build-knowledge of datafolder (or Runtime ProductName)
- Old settings are kept in-place, and not removed
- Commonly used paths are fetched on module-init
- Add copy-constructor to SystemLocal, to circumvent sanitizing already sanitized paths

Format is customized java properties format.
- Allow arbitrary keys with spaces etc. by using `"Key With Space"=Value`
- ~Allows "unsetting" a preference with `Key=`~
  Skips loading old configuration if new configuration exists
  -  ~Required for cases where the legacy settings got overriden/reset to default in runtime~

Resolves #14566


AliveDevil:
Currently missing:
- [ ] In case running in UWP context reflect changed path from `setFactories()` for User-Config file
  - `GetUserConfig()` should return `%LocalAppdata%\Packages\PackageIdentity\…` instead of `%AppData%\DataFolder\…`.

AliveDevil:
Testing revealed:
We cannot, at all, write to %AppData%\Cyberduck if that folder didn't exist prior running Cyberduck as UWP package.
Need to review all paths, that they refer to the correct location depending on UWP or standalone execution.

ylangisc (approved):

PR #14607, opened 2023-05-02
Handle AV verdicts

dkocher:


ylangisc:
Can we add some tests using the EICAR test file?

ylangisc (approved):

PR #14641, opened 2023-05-09
Fix #14639.

### LINKED ISSUE #14639: s3.listing.metadata.enable=true shows wrong timestamp (rclone)
**Describe the bug**

We are using rclone for archiving of large amounts of data and have users browse the archive either with a cli tool or with Cyberduck. We use that special profile called s3 (timestamps) that is supposed to show posix mtime metadata instead of the S3 object creation time using s3.listing.metadata.enable=true . If you are archiving a file that is 5 years old, the object creation time is less relevant and users want to see when the file was last modified (in fact many users do not understand the object time at all) 

Unfortunately, there seems to be an issue with the correct display of mtime metadata.  When I check https://www.unixtimestamp.com/ for that number in mtime I get April 20th 2023 which is correct but Cyberduck with S3 (timestamps) profile shows a time in 1970 
image

The timestamps were created with Rclone but that is perhaps less relevant since Unix time is clearly defined as seconds since 1/1/1970 and it may be a float if you want to include milli or nanoseconds. 

Also, it would be lovely if the timestamp listing could be a simple option in the S3 preferences instead of having to go the detour through the profile settings (I realize there is an additional head request for each object required but the slower performance is not issue for us)

 - OS: Windows 11
 - Version [e.g. 8.5.9]

dkocher:


s3/src/main/java/ch/cyberduck/core/s3/S3AttributesAdapter.java
```diff
@@ -83,7 +83,7 @@ public String getDescription() {
             attributes.setMetadata(metadata);
             if(object.containsMetadata(S3TimestampFeature.METADATA_MODIFICATION_DATE)) {
                 try {
-                    attributes.setModificationDate(Double.valueOf(object.getMetadata(S3TimestampFeature.METADATA_MODIFICATION_DATE).toString()).longValue());
+                    attributes.setModificationDate(Double.valueOf(object.getMetadata(S3TimestampFeature.METADATA_MODIFICATION_DATE).toString()).longValue() * 1000);
```
AliveDevil, line 86: Does this affect how Cyberduck persists Mtime as well?
That is timestamps uploaded through Cyberduck are now off?

dkocher: We are setting the value in millis 🫣 [^1] 

[^1]: https://github.com/iterate-ch/cyberduck/blob/e990d4c43c32e6246bc403b617b45bf0742c524e/s3/src/main/java/ch/cyberduck/core/s3/S3WriteFeature.java#L140-L143

AliveDevil (approved):

ylangisc (approved):

PR #14659, opened 2023-05-11
Fix #14658.

### LINKED ISSUE #14658: Enabling profiles in Preferences does not persist

dkocher:


core/src/main/java/ch/cyberduck/core/ProtocolFactory.java
```diff
@@ -145,7 +145,8 @@ public Local register(final Local file) {
                 log.info(String.format("Register profile %s", profile));
             }
             registered.add(profile);
-            preferences.setProperty(StringUtils.lowerCase(String.format("profiles.%s.%s.enabled", profile.getProtocol(), profile.getProvider())), true);
+            preferences.setProperty(StringUtils.lowerCase(String.format("profiles.%s.%s.enabled",
+                    profile.getIdentifier(), profile.getProvider())), true);
```
ylangisc, line 149: Use either `profile.getIdentifier()` or `profile.getProtocol().getIdentifier()` in `register` and `unregister`.

PR #14662, opened 2023-05-11
Add preference for "Save in Keychain" when unlocking vaults

### LINKED ISSUE #11694: Option to uncheck Save Password by default when unlocking vault
> **@dkocher** created the issue

Regardless of setting in *Preferences → General → Use Keychain*

dkocher:
Fix #11694.


cryptomator/src/main/java/ch/cyberduck/core/cryptomator/CryptoVault.java
```diff
@@ -273,6 +273,7 @@ private void unlock(final VaultConfig vaultConfig, final String passphrase, fina
                     bookmark, LocaleFactory.localizedString("Unlock Vault", "Cryptomator"),
                     message,
                     new LoginOptions()
+                            .keychain(preferences.getBoolean("cryptomator.keychain"))
```
AliveDevil, line 276: We already have "vault.keychain"[^1]
[^1]: https://github.com/iterate-ch/cyberduck/blob/465be7018aeeca4dc2bf6161d9ce83824460004c/cryptomator/src/main/java/ch/cyberduck/core/cryptomator/CryptoVault.java#LL286C75-L286C75

AliveDevil (changes requested):
What's the state here[^1]?
[^1]: https://github.com/iterate-ch/cyberduck/blob/465be7018aeeca4dc2bf6161d9ce83824460004c/core/src/main/java/ch/cyberduck/core/vault/VaultCredentials.java#LL30C50-L30C51

There is conflicting/duplicate usage now on "cryptomator.keychain" and "vault.keychain".

AliveDevil (approved):

ylangisc (approved):

PR #14796, opened 2023-06-14
OIDC+MinIO S3 - Current development stage

### LINKED ISSUE #10917: Expired temporary credentials (STS) should be refreshed automatically
> **884512a** created the issue

Not requiring program restart
___
Attachments

* [`Cyberduck-creds-expired.png`](https://github-cyberduck-assets.s3.amazonaws.com/10917/Cyberduck-creds-expired.png) (23.6 KiB)

### LINKED ISSUE #13377: aws sso based bookmarks are too difficult to use

### LINKED ISSUE #13804: Suport to login using temporary credentials from Security Token Service (STS API) using OpenID Connect web identity

My goals are:

IsmCadi:
This pull request contains our current state with:
- Authentication to Keycloak
- Fetch STS temporary credentials
- Testcontainers with the included docker-compose configuration
- First tests including the testcontainers (are running)
- A first version of the OIDC Cyberduck Profile in the ressource-folder
- Suggested changes from @chenkins implemented (about STS) https://github.com/SheromS/cyberduck/pull/6

- [ ] Resolves #13804
- [ ] Resolves #10917


oauth/src/main/java/ch/cyberduck/core/oauth/OAuth2RequestInterceptor.java
```diff
@@ -115,8 +115,9 @@ public void process(final HttpRequest request, final HttpContext context) throws
             if(log.isInfoEnabled()) {
                 log.info(String.format("Authorizing service request with OAuth2 tokens %s", tokens));
             }
-            request.removeHeaders(HttpHeaders.AUTHORIZATION);
-            request.addHeader(new BasicHeader(HttpHeaders.AUTHORIZATION, String.format("Bearer %s", tokens.getAccessToken())));
+            // TODO if already have the STS credentials, this adds an authorization header causing permission denied (sending OAuth credentials to STS)
```
dkocher, line 118: I see no reason there should be a OAuth handler registered for the connection to the STS service.

profiles/default/S3-OIDC-minio-localhost.cyberduckprofile
```diff
@@ -0,0 +1,58 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
dkocher, line 1: Move this profile to test resources.

profiles/default/S3-OIDC-google.cyberduckprofile
```diff
@@ -0,0 +1,57 @@
...
+        <string>s3-sts-google</string>
+        <key>Bundled</key>
+        <true/>
+        <key>Description</key>
+        <string>S3 STS AssumeRoleWithWebIdentity with Google</string>
+        <key>Default Nickname</key>
+        <string>S3 STS</string>
+        <key>Default Hostname</key>
+        <string>s3.amazonaws.com</string>
+        <key>Regions</key>
```
dkocher, line 32: Do not restrict available regions.

s3/src/main/java/ch/cyberduck/core/sts/AssumeRoleWithWebIdentitySTSCredentialsConfigurator.java
```diff
@@ -0,0 +1,82 @@
...
+        super(trust, key, prompt);
+    }
+
+    @Override
+    public Credentials configure(final Host host) throws LoginFailureException, LoginCanceledException {
+        final Credentials credentials = new Credentials(host.getCredentials());
+
+        final AWSSecurityTokenService service = AWSSecurityTokenServiceClientBuilder
+                .standard()
+                .withEndpointConfiguration(new AwsClientBuilder.EndpointConfiguration(String.format("%s://%s:%s", host.getProtocol().getScheme(), host.getHostname(),
```
dkocher, line 50: Presumably the STS endpoint is different from the storage endpoint and must be configurable in the connection profile.

s3/src/main/java/ch/cyberduck/core/sts/AssumeRoleWithWebIdentitySTSCredentialsConfigurator.java
```diff
@@ -0,0 +1,82 @@
...
+        AssumeRoleWithWebIdentityRequest webIdReq = new AssumeRoleWithWebIdentityRequest()
+                .withWebIdentityToken(credentials.getOauth().getAccessToken());
+//                .withPolicy("policy")
+//                .withRoleArn("consoleAdmin")
+//                .withRoleSessionName("testSession");
+//                .putCustomQueryParameter("Version", "2011-06-15");
+
+        AssumeRoleWithWebIdentityResult result = service.assumeRoleWithWebIdentity(webIdReq);
+        com.amazonaws.services.securitytoken.model.Credentials cred = result.getCredentials();
+        System.out.println(cred.toString());
```
dkocher, line 74: Use logger.

s3/src/main/java/ch/cyberduck/core/sts/AssumeRoleWithWebIdentitySTSCredentialsConfigurator.java
```diff
@@ -0,0 +1,82 @@
...
+                    @Override
+                    public void refresh() {
+
+                    }
+                })
+                .build();
+
+        AssumeRoleWithWebIdentityRequest webIdReq = new AssumeRoleWithWebIdentityRequest()
+                .withWebIdentityToken(credentials.getOauth().getAccessToken());
+//                .withPolicy("policy")
```
dkocher, line 67: Make configurable in connection profile using custom properties.

s3/src/main/java/ch/cyberduck/core/sts/AssumeRoleWithWebIdentitySTSCredentialsConfigurator.java
```diff
@@ -0,0 +1,82 @@
...
+
+    public AssumeRoleWithWebIdentitySTSCredentialsConfigurator(final X509TrustManager trust, final X509KeyManager key, final PasswordCallback prompt) {
+        super(trust, key, prompt);
+    }
+
+    @Override
+    public Credentials configure(final Host host) throws LoginFailureException, LoginCanceledException {
+        final Credentials credentials = new Credentials(host.getCredentials());
+
+        final AWSSecurityTokenService service = AWSSecurityTokenServiceClientBuilder
```
dkocher, line 48: Obtain service using or overriding `getTokenService` with custom client configuration.

s3/src/main/java/ch/cyberduck/core/s3/S3WebIdentityTokenExpiredResponseInterceptor.java
```diff
@@ -0,0 +1,67 @@
+package ch.cyberduck.core.s3;
```
dkocher, line 1: Verify if implementation can be shared with `S3TokenExpiredResponseInterceptor`.

s3/src/main/java/ch/cyberduck/core/s3/S3WebIdentityTokenExpiredResponseInterceptor.java
```diff
@@ -0,0 +1,67 @@
...
+    public S3WebIdentityTokenExpiredResponseInterceptor(final S3Session session, final X509TrustManager trust, final X509KeyManager key, final LoginCallback prompt) {
+        this.host = session.getHost();
+        this.configurator = new AssumeRoleWithWebIdentitySTSCredentialsConfigurator(trust, key, prompt);
+    }
+
+    @Override
+    public boolean retryRequest(final HttpResponse response, final int executionCount, final HttpContext context) {
+        if(executionCount <= MAX_RETRIES) {
+            switch(response.getStatusLine().getStatusCode()) {
+                case HttpStatus.SC_FORBIDDEN:
```
dkocher, line 50: Verify this is indeed `403` and not `400`

s3/src/main/java/ch/cyberduck/core/sts/STSCredentialsConfigurator.java
```diff
@@ -248,7 +248,7 @@ else if(!profiles.containsKey(basicProfile.getRoleSourceProfile())) {
                     log.debug(String.format("Configure credentials from basic profile %s", basicProfile.getProfileName()));
                 }
                 final Map<String, String> profileProperties = basicProfile.getProperties();
-                if(profileProperties.containsKey("sso_start_url") || profileProperties.containsKey("sso_session")) {
+                if(profileProperties.containsKey("sso_start_url")) {
```
dkocher, line 251: What is the reasoning for this change which seems unrelated. This applies to connections using STS  tokens obtained from cache of AWS CLI.

s3/src/main/java/ch/cyberduck/core/s3/S3OidcSession.java
```diff
@@ -0,0 +1,64 @@
...
+import ch.cyberduck.core.ssl.ThreadLocalHostnameDelegatingTrustManager;
+import ch.cyberduck.core.ssl.X509KeyManager;
+import ch.cyberduck.core.ssl.X509TrustManager;
+import ch.cyberduck.core.threading.CancelCallback;
+
+import org.apache.http.impl.client.HttpClientBuilder;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+public class S3OidcSession extends S3Session {
```
dkocher, line 33: Any usage for this class?

oauth/src/main/java/ch/cyberduck/core/oauth/OAuth2RequestInterceptor.java
IsmCadi: You are right, we are still working on how to use the requestInterceptor only for keylcoak and not for STS

s3/src/main/java/ch/cyberduck/core/sts/AssumeRoleWithWebIdentitySTSCredentialsConfigurator.java
SheromS: Thank you for the hint. We need to consider this because with other provider than MinIO it could be at different endpoints.

s3/src/main/java/ch/cyberduck/core/sts/AssumeRoleWithWebIdentitySTSCredentialsConfigurator.java
dkocher: All possible configuration properties for the request that may be required.

s3/src/main/java/ch/cyberduck/core/s3/S3WebIdentityTokenExpiredResponseInterceptor.java
SheromS: We verified it by testing and with the MinIO documentation that is is 403 Forbidden if the credentials are expired.

sebastiangraf:
A short summary and a  small request from my side:
Summary:
The programming part is entirely covered by this PR. Upon merge, we start to finalize the project by writing the report. Aim is the closure of this PR at the 10.7.23.
From a functional side and a higher level, the following components are necessary (and tbh: I do not know if they are already included in this PR or not but want to write them down for transparency and reference:

- All settings regarding the backends (oidc, sts, storage) must be configurable via a profile. No information should be hardcoded.
- The creation of a test profile containing test users is ok
- the tests should include two oidc providers to be confident that the implementation is agnostic with respect to the provider. The same tests might be executed multiple times with different profiles
- not PR from external work must be merged in this sourcebranch. However, if a PR is provided with some help, it is up to @IsmCadi and @SheromS to decide to include it or not (if it would be helpful).

And now the request:
@dkocher : Could you list transparently what is necessary in this PR to be merged so we can work towards a quick resolution of this PR? Thanks.

.github/workflows/build.yml
```diff
@@ -1,6 +1,7 @@
 name: Unit Tests
 
 on:
+  workflow_dispatch:
```
dkocher, line 4: Revert change.

s3/src/main/java/ch/cyberduck/core/s3/S3WebIdentityTokenExpiredResponseInterceptor.java
dkocher: And for AWS?

s3/src/main/java/ch/cyberduck/core/s3/S3WebIdentityTokenExpiredResponseInterceptor.java
```diff
@@ -0,0 +1,85 @@
...
+                    try {
+                        session.refreshOAuthTokens();
+                        Credentials credentials = configurator.configure(host);
+                        session.getClient().setProviderCredentials(credentials.isAnonymousLogin() ? null :
+                                new AWSSessionCredentials(credentials.getUsername(), credentials.getPassword(),
+                                        credentials.getToken()));
+                        System.out.println(session.getClient().getProviderCredentials().getAccessKey());
```
dkocher, line 67: Use logger.

s3/src/main/java/ch/cyberduck/core/sts/AssumeRoleWithWebIdentitySTSCredentialsConfigurator.java
```diff
@@ -0,0 +1,108 @@
+package ch.cyberduck.core.sts;/*
```
dkocher, line 1: Formatting.

s3/src/main/java/ch/cyberduck/core/sts/AssumeRoleWithWebIdentitySTSCredentialsConfigurator.java
```diff
@@ -0,0 +1,108 @@
...
+import com.amazonaws.auth.profile.internal.BasicProfile;
+import com.amazonaws.client.builder.AwsClientBuilder;
+import com.amazonaws.regions.Regions;
+import com.amazonaws.services.securitytoken.AWSSecurityTokenService;
+import com.amazonaws.services.securitytoken.AWSSecurityTokenServiceClientBuilder;
+import com.amazonaws.services.securitytoken.model.AssumeRoleWithWebIdentityRequest;
+import com.amazonaws.services.securitytoken.model.AssumeRoleWithWebIdentityResult;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class AssumeRoleWithWebIdentitySTSCredentialsConfigurator extends AWSProfileSTSCredentialsConfigurator {
```
dkocher, line 51: Extract common base class for existing `AWSProfileSTSCredentialsConfigurator` and new implemenation.

s3/src/main/java/ch/cyberduck/core/sts/AssumeRoleWithWebIdentitySTSCredentialsConfigurator.java
```diff
@@ -0,0 +1,108 @@
...
+        credentials.setToken(cred.getSessionToken());
+
+        return credentials;
+    }
+
+    @Override
+    public AWSSecurityTokenService getTokenService(final Host host, final String region, final String accessKey, final String secretKey, final String sessionToken) {
+        return AWSSecurityTokenServiceClientBuilder
+                .standard()
+                .withEndpointConfiguration(new AwsClientBuilder.EndpointConfiguration(String.format("%s://%s:%s", host.getProtocol().getScheme(), host.getHostname(),
```
dkocher, line 93: Most certainly a different STS endpoint than the bookmark configuration. Must be configurable in connection profile.

s3/src/main/java/ch/cyberduck/core/sts/AssumeRoleWithWebIdentitySTSCredentialsConfigurator.java
```diff
@@ -0,0 +1,108 @@
...
+
+        credentials.setUsername(cred.getAccessKeyId());
+        credentials.setPassword(cred.getSecretAccessKey());
+        credentials.setToken(cred.getSessionToken());
+
+        return credentials;
+    }
+
+    @Override
+    public AWSSecurityTokenService getTokenService(final Host host, final String region, final String accessKey, final String secretKey, final String sessionToken) {
```
dkocher, line 90: Share implementation with existing `AWSProfileSTSCredentialsConfigurator` implementation.

s3/src/test/java/ch/cyberduck/core/oidc/testenv/OidcAuthenticationTest.java
```diff
@@ -0,0 +1,102 @@
+package ch.cyberduck.core.oidc.testenv;/*
```
dkocher, line 1: Formatting.

s3/src/main/java/ch/cyberduck/core/s3/S3Session.java
```diff
@@ -269,6 +313,11 @@ public static boolean isAwsHostname(final String hostname, boolean cn) {
         return hostname.matches("([a-z0-9\\-]+\\.)?s3(\\.dualstack)?(\\.[a-z0-9\\-]+)?(\\.vpce)?\\.amazonaws\\.com");
     }
 
+    public void refreshOAuthTokens() throws BackgroundException {
```
dkocher, line 316: Move to `S3WebIdentityTokenExpiredResponseInterceptor`

dkocher (changes requested):
- [x] Tests ✅ 
- [ ] Resolve TODOs
- [ ] Resolve review comments
- [ ] Squash commits

s3/src/main/java/ch/cyberduck/core/sts/STSSecurityTokenHeaderHttpRequestInterceptor.java
```diff
@@ -0,0 +1,43 @@
...
+    private final S3Session session;
+    private static final String securityTokenHeaderKey = "x-amz-security-token";
+
+    public STSSecurityTokenHeaderHttpRequestInterceptor(S3Session session) {
+        this.session = session;
+    }
+
+    @Override
+    public void process(final org.apache.http.HttpRequest request, final HttpContext context) {
+        request.setHeader(securityTokenHeaderKey, ((AWSSessionCredentials) session.getClient().getProviderCredentials()).getSessionToken());
```
SheromS, line 41: updates the x-amz-security-token header but still there are hashes which need to be updated somehow.

SheromS: this includes headers like "Authorization" or "x-amz-content-sha"

.github/workflows/build.yml
IsmCadi: done

s3/src/main/java/ch/cyberduck/core/s3/S3WebIdentityTokenExpiredResponseInterceptor.java
IsmCadi: Done.

s3/src/main/java/ch/cyberduck/core/sts/AssumeRoleWithWebIdentitySTSCredentialsConfigurator.java
IsmCadi: Done.

s3/src/test/java/ch/cyberduck/core/oidc/testenv/OidcAuthenticationTest.java
IsmCadi: Done.

core/src/main/java/ch/cyberduck/core/Profile.java
```diff
@@ -536,6 +536,57 @@ public boolean isOAuthPKCE() {
         return this.bool("OAuth PKCE");
     }
 
+    @Override
```
dkocher, line 539: As previously discussed in https://github.com/iterate-ch/cyberduck/pull/14796/files/36d33a7a4e0b2aeb52b33ae3b762c75cff23e338#r1229312937 please use custom properties instead in the profile and set defaults if required in `defaults.properties`.

dkocher (changes requested):

s3/src/main/java/ch/cyberduck/core/s3/S3WebIdentityTokenExpiredResponseInterceptor.java
SheromS: Error handling has been changed. Check the latest state.

core/src/main/java/ch/cyberduck/core/Profile.java
SheromS: Good idea. It is to clarify where the definition of the s3.bucket.virtualhost.disable is? @chenkins used it in the Properties section also. 
<key>Properties</key>
        <array>
            <string>s3.bucket.virtualhost.disable=true</string>
        </array>
        
In my opinion this works only if we can garantee that the Properties are only used for the request parameters:
//                .withRoleArn("consoleAdmin")
//                .withRoleSessionName("testSession");
//                .withDurationSeconds(900);

dkocher: A default definition can be added in `default.properties` [^1] if applicable and provided in the *.cyberduckprofile files for specific services.

[^1]: https://github.com/iterate-ch/cyberduck/blob/master/defaults/src/main/resources/default.properties

SheromS: Thanks. We set the default.properties. We don't know how to overwrite the default values by the custom properties.
if (host.getProtocol().getProperties().containsKey("s3.assumerole.durationseconds")) {
            System.out.println("set duration second with profile value");
            webIdReq = webIdReq.withDurationSeconds(host.getProtocol().getProperties().get("s3.assumerole.durationseconds"));
        }
        else {
            webIdReq = webIdReq.withDurationSeconds(new HostPreferences(host).getInteger("s3.assumerole.durationseconds"));
        }

dkocher: Just use `new HostPreferences(host).getInteger("s3.assumerole.durationseconds")` which implements the fallback.

1. Search in properties saved with bookmark
2. Search in properties in connection profile
3. Search in application preferences

s3/src/main/java/ch/cyberduck/core/sts/AssumeRoleWithWebIdentitySTSCredentialsConfigurator.java
```diff
@@ -48,13 +54,51 @@ public Credentials configure(final Host host) {
...
         AssumeRoleWithWebIdentityRequest webIdReq = new AssumeRoleWithWebIdentityRequest()
                 .withWebIdentityToken(credentials.getOauth().getAccessToken());
-//                .withPolicy("policy")
-//                .withRoleArn("consoleAdmin")
-//                .withRoleSessionName("testSession");
-//                .putCustomQueryParameter("Version", "2011-06-15");
 
-        if(log.isDebugEnabled()) {
-            log.debug("Assuming role with web identity for host: {}", host);
+        String durationsecondsProp = "s3.assumerole.durationseconds";
```
dkocher, line 58: Use `HostPreferences` and override properties in .cyberduckprofile.

s3/src/main/java/ch/cyberduck/core/sts/AssumeRoleWithWebIdentitySTSCredentialsConfigurator.java
SheromS: You mean to create a new abstract class and both classes extend this abstract class? I was just tryting to do this. In my opinion the abstract class would only contain the logger, the trust- and key-manager. The code is very different in these two classes. Is that what you want?

dkocher: It should be possible to share the token service from `getTokenService`.

s3/src/main/java/ch/cyberduck/core/sts/AssumeRoleWithWebIdentitySTSCredentialsConfigurator.java
```diff
@@ -0,0 +1,141 @@
...
+            webIdReq = webIdReq.withRoleSessionName(host.getProperty(rolesessionnameProp));
+        }
+        else if(host.getProtocol().getProperties().containsKey(rolesessionnameProp)) {
+            webIdReq = webIdReq.withRoleSessionName(host.getProtocol().getProperties().get(rolesessionnameProp));
+        }
+        else if(StringUtils.isNotBlank(PreferencesFactory.get().getProperty(rolesessionnameProp))) {
+            webIdReq = webIdReq.withRoleSessionName(new HostPreferences(host).getProperty(rolesessionnameProp));
+        }
+
+        AssumeRoleWithWebIdentityResult result = service.assumeRoleWithWebIdentity(webIdReq);
```
dkocher, line 104: Missing exception handling for failures such as `InvalidIdentityTokenException`.

dkocher (changes requested):
- [ ] Should the session token from STS be refreshed preemptively prior expiry? The expiry is known from the `Expiration` set in the `AssumeRoleWithWebIdentityResponse ` response model from `AssumeRoleWithWebIdentity` API.

s3/src/main/java/ch/cyberduck/core/sts/AssumeRoleWithWebIdentitySTSCredentialsConfigurator.java
SheromS: I will create a new base class containing the following fields. Is that what you mean? The getTokenService-Method differs quite a lot between both classes so their implementation will be in the subclasses.

public abstract class AbstractSTSCredentialsConfigurator implements CredentialsConfigurator {
    private static final Logger log = LogManager.getLogger(AbstractSTSCredentialsConfigurator.class);

    protected final X509TrustManager trust;
    protected final X509KeyManager key;
    protected AWSSecurityTokenService service;

core/src/main/java/ch/cyberduck/core/Profile.java
SheromS: This is now implemented according to your feedback.

s3/src/main/java/ch/cyberduck/core/sts/AssumeRoleWithWebIdentitySTSCredentialsConfigurator.java
SheromS: is our code correct now?
Now we also check if durationSeconds is a valid size (900s - 42300s). If it is out of bounds I throw a IllegalArgumentException. Is that fine?

s3/src/main/java/ch/cyberduck/core/s3/S3WebIdentityTokenExpiredResponseInterceptor.java
```diff
@@ -0,0 +1,131 @@
...
+                    final S3ServiceException failure;
+                    if(null != response.getEntity()) {
+                        EntityUtils.updateEntity(response, new BufferedHttpEntity(response.getEntity()));
+                        failure = new S3ServiceException(response.getStatusLine().getReasonPhrase(),
+                                EntityUtils.toString(response.getEntity()));
+                    }
+                    // minio sometimes packs the error code and description in the http header
+                    else{
+                        failure = new S3ServiceException(response.getStatusLine().getReasonPhrase());
+                        if(response.containsHeader("x-minio-error-code")) {
```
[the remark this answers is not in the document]
SheromS: Unfortunately, minio does not always send the error details in the response body ( `response.getEntity == null` ). In this scenario the only error description is found in these minio specific headers.

s3/src/main/java/ch/cyberduck/core/sts/AssumeRoleWithWebIdentitySTSCredentialsConfigurator.java
SheromS: There are a whole bunch of possible exceptions. What is the right reaction in the catch-statements?
`InvalidIdentityTokenException` -> oAuth-authorize
`ExpiredTokenException` -> oAuth-refresh
other Exception -> log error message?

s3/src/main/java/ch/cyberduck/core/s3/S3WebIdentityTokenExpiredResponseInterceptor.java
```diff
@@ -0,0 +1,131 @@
...
+            if(log.isWarnEnabled()) {
+                log.warn(String.format("Skip retry for response %s after %d executions", response, executionCount));
+            }
+        }
+        return false;
+    }
+
+    private void refreshOAuthAndSTS() {
+        try {
+            if(host.getCredentials().getOauth().isExpired()) {
```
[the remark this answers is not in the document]
SheromS: Good point. Is it a good idea to save the expiry time as a host property? Or do we I access this information after it is reset?

s3/src/main/java/ch/cyberduck/core/sts/AssumeRoleWithWebIdentitySTSCredentialsConfigurator.java
SheromS: Is now implemented like this

dkocher (changes requested):
Please rebase on #14882. This should allow to properly sign requests after refreshing tokens.

Because the `CredentialsConfigurator` assumes configuration of credentials from static resources such as OpenSSH configuration directives or AWS CLI configuration files, it does not declare exception to be thrown and in general does not seem to be suited for this use case. 

We should rather extend or make use of `OAuth2RequestInterceptor` to handle STS on top. This will allow to 
* Refresh OAuth tokens if not already preemptively done in handler
* Preemptively refresh STS token when expired prior sending the request
* Allow to propagate exceptions from the STS API calls when invoking in `login` in the same way as `OAuth2AuthorizationService#authorize`

- [ ] Make sure we have test coverage for expired STS tokens.
- [ ] Resolves TODOs
- [ ] Resolve comments

oauth/src/main/java/ch/cyberduck/core/oauth/OAuth2AuthorizationService.java
```diff
@@ -135,25 +136,26 @@ public OAuthTokens authorize(final Host bookmark, final LoginCallback prompt, fi
         if(log.isDebugEnabled()) {
             log.debug(String.format("Start new OAuth flow for %s with missing access token", bookmark));
         }
-        final TokenResponse response;
+        // Save access key and refresh key
         switch(flowType) {
             case AuthorizationCode:
-                response = this.authorizeWithCode(bookmark, prompt);
-                break;
+                final IdTokenResponse authorizationCodeResponse = this.authorizeWithCode(bookmark, prompt);
```
dkocher, line 142: Extend `OAuthTokens` instead with field for `id_token` from OIDC.

s3/src/main/java/ch/cyberduck/core/sts/AssumeRoleWithWebIdentitySTSCredentialsConfigurator.java
```diff
@@ -43,12 +45,17 @@ public AssumeRoleWithWebIdentitySTSCredentialsConfigurator(final X509TrustManage
...
     public Credentials configure(final Host host) {
-        final Credentials credentials = new Credentials(host.getCredentials());
+        final Credentials hostCredentials = new Credentials(host.getCredentials());
 
         service = this.getTokenService(host, null, null, null, null);
 
-        AssumeRoleWithWebIdentityRequest webIdReq = new AssumeRoleWithWebIdentityRequest()
-                .withWebIdentityToken(credentials.getOauth().getAccessToken());
+        AssumeRoleWithWebIdentityRequest webIdReq = new AssumeRoleWithWebIdentityRequest();
+        if(hostCredentials.getToken().isEmpty()) {
```
dkocher, line 53: The OAuth token should be obtained from the `OAuth2RequestInterceptor` which stores the current values (to make sure you get the updated token values after a refresh.

dkocher (changes requested):
- [ ] Obtain OAuth tokens from `OAuth2RequestInterceptor` for AssumeRoleWithWebIdenttiy

core/src/main/java/ch/cyberduck/core/exception/WebIdentityTokenExpiredException.java
```diff
@@ -0,0 +1,28 @@
+package ch.cyberduck.core.exception;/*
```
dkocher, line 1: Is reusing `ExpiredTokenException` not suitable?

s3/src/main/java/ch/cyberduck/core/sts/STSCredentialsRequestInterceptor.java
```diff
@@ -0,0 +1,185 @@
...
+        }
+        catch(AWSSecurityTokenServiceException e) {
+            log.error(e.getErrorMessage());
+        }
+        return credentials;
+    }
+
+    private AWSSecurityTokenService getTokenService(final Host host, final String region, final String accessKey, final String secretKey, final String sessionToken) {
+        final ClientConfiguration configuration = new CustomClientConfiguration(host,
+                new ThreadLocalHostnameDelegatingTrustManager(trust, host.getHostname()), key);
```
dkocher, line 143: Trust manager should be configured with STS hostname.

s3/src/main/java/ch/cyberduck/core/sts/STSCredentialsRequestInterceptor.java
```diff
@@ -0,0 +1,185 @@
...
+                                            PasswordCallback prompt, S3Session session) {
+        super(client, host);
+        this.trust = trust;
+        this.key = key;
+        this.prompt = prompt;
+        this.session = session;
+    }
+
+    @Override
+    public void process(final org.apache.http.HttpRequest request, final HttpContext context) throws IOException {
```
dkocher, line 71: Duplicated code from OAuth Handler.

dkocher (changes requested):
- [ ] Please validate your refresh semantics with tests
- [ ] Make sure request and error handlers are configured in a cascading manner that OAuth is handled prior STS

core/src/main/java/ch/cyberduck/core/exception/WebIdentityTokenExpiredException.java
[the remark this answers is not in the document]
IsmCadi: Yes, it makes sense to use this class also.

s3/src/main/java/ch/cyberduck/core/s3/S3WebIdentityTokenExpiredResponseInterceptor.java
```diff
@@ -0,0 +1,131 @@
...
+        this.host = session.getHost();
+        this.configurator = new AssumeRoleWithWebIdentitySTSCredentialsConfigurator(trust, key, prompt);
+        this.authorizationService = authorizationService;
+        this.prompt = prompt;
+    }
+
+    @Override
+    public boolean retryRequest(final HttpResponse response, final int executionCount, final HttpContext context) {
+        if(executionCount <= MAX_RETRIES) {
+            if (400 <= response.getStatusLine().getStatusCode() && response.getStatusLine().getStatusCode() < 500) {
```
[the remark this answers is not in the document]
SheromS: we have changed it to catch 400 and 403 Status Codes. Usually it is a 400 when the token is expired. If a request leaves the app with an expired AccessKeyId then we get a 403 and are able to react with a reauthorization of oAuth.

dkocher:
Superseded by #14893.

PR #14893, opened 2023-07-09
Connect to Minio with temporary credentials from STS obtained with OpenID Connect token from Keycloak

### LINKED ISSUE #10917: Expired temporary credentials (STS) should be refreshed automatically

### LINKED ISSUE #13377: aws sso based bookmarks are too difficult to use

### LINKED ISSUE #13804: Suport to login using temporary credentials from Security Token Service (STS API) using OpenID Connect web identity

SheromS:
This pull request contains our current state with:

- Authentication to Keycloak
- Fetch STS temporary credentials
- Testcontainers with the included docker-compose configuration
- First tests including the testcontainers (are running)
- A first version of the OIDC Cyberduck Profile in the ressource-folder
- Suggested changes from @chenkins implemented (about STS) https://github.com/SheromS/cyberduck/pull/6
- rebase on [#14882](https://github.com/iterate-ch/cyberduck/pull/14882). This allows to properly sign requests after refreshing tokens.

- [x] Resolves #13804
- [ ] Resolves #10917

- [x] Make sure we have test coverage for expired STS tokens.
- [x] Resolves TODOs
- [x] Resolve comments

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,12 +31,3 @@ jobs:
       - name: Publish Test Report
         if: ${{ always() }}
         uses: ScaCap/action-surefire-report@v1.7.0
-
-
-services:
-  ryuk:
-    image: testcontainers/ryuk:0.3.0
-    options: --network=host
-
-variables:
-  TESTCONTAINERS_HOST_OVERRIDE: "host.docker.internal"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,3 +31,12 @@ jobs:
       - name: Publish Test Report
         if: ${{ always() }}
         uses: ScaCap/action-surefire-report@v1.7.0
+
+
+services:
+  ryuk:
+    image: testcontainers/ryuk:0.3.0
+    options: --network=host
+
+variables:
+  TESTCONTAINERS_HOST_OVERRIDE: "host.docker.internal"
--- a/core/src/main/java/ch/cyberduck/core/DefaultHostPasswordStore.java
+++ b/core/src/main/java/ch/cyberduck/core/DefaultHostPasswordStore.java
@@ -161,7 +161,9 @@ public abstract class DefaultHostPasswordStore implements HostPasswordStore {
                             String.format("%s OAuth2 Access Token", prefix)),
                     this.getPassword(bookmark.getProtocol().getScheme(), bookmark.getPort(), hostname,
                             String.format("%s OAuth2 Refresh Token", prefix)),
-                    expiry != null ? Long.parseLong(expiry) : -1L));
+                    expiry != null ? Long.parseLong(expiry) : -1L,
+                    this.getPassword(bookmark.getProtocol().getScheme(), bookmark.getPort(), hostname,
+                            String.format("%s OIDC Id Token", prefix))));
         }
         catch(LocalAccessDeniedException e) {
             log.warn(String.format("Failure %s searching in keychain", e));
@@ -232,6 +234,11 @@ public abstract class DefaultHostPasswordStore implements HostPasswordStore {
                 this.addPassword(this.getOAuthHostname(bookmark), String.format("%s OAuth2 Token Expiry", prefix),
                         String.valueOf(credentials.getOauth().getExpiryInMilliseconds()));
             }
+            if(StringUtils.isNotBlank(credentials.getOauth().getIdToken())) {
+                this.addPassword(bookmark.getProtocol().getScheme(),
+                        bookmark.getPort(), this.getOAuthHostname(bookmark),
+                        String.format("%s OIDC Id Token", prefix), credentials.getOauth().getRefreshToken());
+            }
         }
     }
 
@@ -271,6 +278,10 @@ public abstract class DefaultHostPasswordStore implements HostPasswordStore {
             // Save expiry
             if(credentials.getOauth().getExpiryInMilliseconds() != null) {
                 this.deletePassword(this.getOAuthHostname(bookmark), String.format("%s OAuth2 Token Expiry", prefix));
+            }
+            if(StringUtils.isNotBlank(credentials.getOauth().getRefreshToken())) {
+                this.deletePassword(protocol.getScheme(), bookmark.getPort(), this.getOAuthHostname(bookmark),
+                        String.format("%s OIDC Id Token", prefix));
             }
         }
     }
--- a/core/src/main/java/ch/cyberduck/core/DefaultHostPasswordStore.java
+++ b/core/src/main/java/ch/cyberduck/core/DefaultHostPasswordStore.java
@@ -279,7 +279,7 @@ public abstract class DefaultHostPasswordStore implements HostPasswordStore {
             if(credentials.getOauth().getExpiryInMilliseconds() != null) {
                 this.deletePassword(this.getOAuthHostname(bookmark), String.format("%s OAuth2 Token Expiry", prefix));
             }
-            if(StringUtils.isNotBlank(credentials.getOauth().getRefreshToken())) {
+            if(StringUtils.isNotBlank(credentials.getOauth().getIdToken())) {
                 this.deletePassword(protocol.getScheme(), bookmark.getPort(), this.getOAuthHostname(bookmark),
                         String.format("%s OIDC Id Token", prefix));
             }
--- a/core/src/main/java/ch/cyberduck/core/DefaultHostPasswordStore.java
+++ b/core/src/main/java/ch/cyberduck/core/DefaultHostPasswordStore.java
@@ -237,7 +237,7 @@ public abstract class DefaultHostPasswordStore implements HostPasswordStore {
             if(StringUtils.isNotBlank(credentials.getOauth().getIdToken())) {
                 this.addPassword(bookmark.getProtocol().getScheme(),
                         bookmark.getPort(), this.getOAuthHostname(bookmark),
-                        String.format("%s OIDC Id Token", prefix), credentials.getOauth().getRefreshToken());
+                        String.format("%s OIDC Id Token", prefix), credentials.getOauth().getIdToken());
             }
         }
     }
--- a/oauth/src/main/java/ch/cyberduck/core/oauth/OAuth2AuthorizationService.java
+++ b/oauth/src/main/java/ch/cyberduck/core/oauth/OAuth2AuthorizationService.java
@@ -51,7 +51,6 @@ import com.google.api.client.auth.oauth2.ClientParametersAuthentication;
 import com.google.api.client.auth.oauth2.Credential;
 import com.google.api.client.auth.oauth2.PasswordTokenRequest;
 import com.google.api.client.auth.oauth2.RefreshTokenRequest;
-import com.google.api.client.auth.oauth2.TokenResponse;
 import com.google.api.client.auth.oauth2.TokenResponseException;
 import com.google.api.client.auth.openidconnect.IdTokenResponse;
 import com.google.api.client.http.GenericUrl;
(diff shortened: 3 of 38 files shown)


dkocher:
Somehow this is based on 10045f45 from bugfix/GH-10917-authentication-interceptor but contains newer commits from `master`. Please rebase again.

s3/src/test/resources/S3-OIDC-Testing.cyberduckprofile
```diff
@@ -0,0 +1,67 @@
...
+        <key>STS Endpoint</key>
+        <string>http://localhost:9000</string>
+        <key>Scopes</key>
+        <array>
+            <string>openid</string>
+            <string>profile</string>
+            <string>minio-authorization</string>
+        </array>
+        <key>Password Configurable</key>
+        <false/>
```
dkocher, line 57: Because `PasswordGrant` is used, this should be `true`.

s3/src/test/resources/S3-OIDC-Testing.cyberduckprofile
```diff
@@ -0,0 +1,67 @@
...
+        <key>Username Configurable</key>
+        <false/>
```
dkocher, line 59: Because `PasswordGrant` is used, this should be `true`.

s3/src/main/java/ch/cyberduck/core/s3/S3WebIdentityTokenExpiredResponseInterceptor.java
```diff
@@ -0,0 +1,126 @@
...
+            if(log.isWarnEnabled()) {
+                log.warn(String.format("Skip retry for response %s after %d executions", response, executionCount));
+            }
+        }
+        return false;
+    }
+
+    private void refreshOAuthAndSTS() {
+        try {
+            if(host.getCredentials().getOauth().isExpired()) {
```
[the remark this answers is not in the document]
dkocher: Either always refresh the OAuth tokens or make it dependent on the error response code if possible.

s3/src/main/java/ch/cyberduck/core/sts/STSCredentialsRequestInterceptor.java
```diff
@@ -0,0 +1,206 @@
...
+                        this.save(this.refresh(tokens));
+                    }
+                    catch(InteroperabilityException | LoginFailureException e3) {
+                        log.warn(String.format("Failure %s refreshing OAuth tokens", e3));
+                        try {
+                            this.save(this.authorize(host, prompt, new DisabledCancelCallback()));
+                        }
+                        catch(BackgroundException e) {
+                            log.warn(String.format("Failure %s OAuth authentication", e));
+                            throw new BackgroundException(e);
```
dkocher, line 89: Why is `e` re-wrapped?

s3/src/main/java/ch/cyberduck/core/sts/STSCredentialsRequestInterceptor.java
```diff
@@ -0,0 +1,206 @@
...
+        this.key = key;
+        this.prompt = prompt;
+        this.session = session;
+    }
+
+    @Override
+    public void process(final org.apache.http.HttpRequest request, final HttpContext context) throws HttpException, IOException {
+        if(System.currentTimeMillis() >= stsExpiryInMilliseconds) {
+            try {
+                if(tokens.isExpired()) {
```
dkocher, line 78: It should be possible to just call `super.process` here to refresh the OAuth tokens.

s3/src/main/java/ch/cyberduck/core/sts/STSCredentialsRequestInterceptor.java
```diff
@@ -0,0 +1,206 @@
...
+        return credentials;
+    }
+
+    private AWSSecurityTokenService getTokenService(final Host host) {
+        final ClientConfiguration configuration = new CustomClientConfiguration(host,
+                new ThreadLocalHostnameDelegatingTrustManager(trust, host.getProtocol().getSTSEndpoint()), key);
+        return AWSSecurityTokenServiceClientBuilder
+                .standard()
+                .withEndpointConfiguration(new AwsClientBuilder.EndpointConfiguration(host.getProtocol().getSTSEndpoint(), null))
+                .withCredentials(new AWSCredentialsProvider() {
```
dkocher, line 164: Replace with `new StaticCredentialsProvider(new AnonymousAWSCredentials())`

dkocher (changes requested):
- [ ] Review comments.
- [x] Please run code formatting for all new classes to match existing code style.
- [x] Change package name of test classes to `sts`?

oauth/src/main/java/ch/cyberduck/core/oauth/OAuth2AuthorizationService.java
```diff
@@ -135,25 +136,27 @@ public OAuthTokens authorize(final Host bookmark, final LoginCallback prompt, fi
         if(log.isDebugEnabled()) {
             log.debug(String.format("Start new OAuth flow for %s with missing access token", bookmark));
         }
-        final TokenResponse response;
+        // Save access key and refresh key
         switch(flowType) {
             case AuthorizationCode:
-                response = this.authorizeWithCode(bookmark, prompt);
-                break;
+                final IdTokenResponse authorizationCodeResponse = this.authorizeWithCode(bookmark, prompt);
```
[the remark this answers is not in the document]
SheromS: good point. I added the id token on every occurence and used it in tests `getIdToken()`. I also added it in the class `DefaultHostPasswordStore`. I hope this is good.

s3/src/main/java/ch/cyberduck/core/s3/S3WebIdentityTokenExpiredResponseInterceptor.java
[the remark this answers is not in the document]
SheromS: Now it always refreshes the token in case of these specific s3 error codes which makes sense.

s3/src/main/java/ch/cyberduck/core/s3/S3WebIdentityTokenExpiredResponseInterceptor.java
```diff
@@ -0,0 +1,126 @@
...
+                try {
+                    final S3ServiceException failure;
+                    if(null != response.getEntity()) {
+                        EntityUtils.updateEntity(response, new BufferedHttpEntity(response.getEntity()));
+                        failure = new S3ServiceException(response.getStatusLine().getReasonPhrase(),
+                                EntityUtils.toString(response.getEntity()));
+                    }
+                    // minio sometimes packs the error code and description in the http header
+                    else{
+                        failure = new S3ServiceException(response.getStatusLine().getReasonPhrase());
```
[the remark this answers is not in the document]
SheromS: The InvalidAccessKeyId error code has no response body.
I made a commented [test](https://github.com/SheromS/cyberduck/blob/9eb90698c2c46557fb48139fc4e17266bfbd749b/s3/src/test/java/ch/cyberduck/core/oidc_sts/OidcAuthenticationTest.java#L151C11-L151C11) which shows an edge case scenario. The request leaves the app right before the expiry time is reached. Because of the network latency the expiry time is reach at the minio service. The AccessKey got just outdated and the Failure gets handled in this class `S3WebIdentityTokenExpiredResponseInterceptor`.
What are your thoughts about that?

s3/src/main/java/ch/cyberduck/core/sts/STSCredentialsRequestInterceptor.java
SheromS: To me this is not possible because in the super refresh there is no catch for the specific exceptions to do an oAuth authorization afterwards. Additionally, it sets a wrong AUTHORIZATION Header which is not a problem because this Header is recalculated in an interceptor later in the chain but to me it is not nice.

SheromS:
- Code formatting for all new classes is done.
- Currently the name of the testing package is oidc_sts. If you wish I change it to sts but in `OidcAuthenticationTest` there are some tests only considering the oidc part.

dkocher (changes requested):
- [x] Change package name of new STS related production and test classes to `ch.cyberduck.core.sts`. While OIDC is part of the specific flow, this could be replaced by other `assumeRole` implementations using LDAP or other identity providers.

oauth/src/main/java/ch/cyberduck/core/oauth/OAuth2AuthorizationService.java
```diff
@@ -136,21 +135,24 @@ public OAuthTokens authorize(final Host bookmark, final LoginCallback prompt, fi
...
+        // Save access token, refresh token and id token
         switch(flowType) {
             case AuthorizationCode:
-                final IdTokenResponse authorizationCodeResponse = this.authorizeWithCode(bookmark, prompt);
-                return credentials.withToken(authorizationCodeResponse.getIdToken()).withOauth(new OAuthTokens(
-                        authorizationCodeResponse.getAccessToken(), authorizationCodeResponse.getRefreshToken(),
-                        null == authorizationCodeResponse.getExpiresInSeconds() ? System.currentTimeMillis() :
-                                System.currentTimeMillis() + authorizationCodeResponse.getExpiresInSeconds() * 1000, authorizationCodeResponse.getIdToken()))
+                response = this.authorizeWithCode(bookmark, prompt);
+                return credentials.withToken(response.getIdToken()).withOauth(new OAuthTokens(
```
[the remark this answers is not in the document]
SheromS: You are right. This is not required anymore.

s3/src/main/java/ch/cyberduck/core/sts/STSCredentialsRequestInterceptor.java
SheromS: I definitely see the point of avoiding code duplication. I don't see how to solve that with the `OAuth2ErrorResponseInterceptor`. This is an `ServiceUnavailableRetryStrategy` and reacts on 401 SC Response but we decided to refresh the tokens preemptively in an interceptor.
I thought about extracting the code to a separate method but there are always tiny differences that lead to incompatibility.

s3/src/main/java/ch/cyberduck/core/s3/S3WebIdentityTokenExpiredResponseInterceptor.java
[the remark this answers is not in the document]
SheromS: We took your [changes](https://github.com/iterate-ch/cyberduck/pull/14893/files/3528d70556808a5c1c61ad2bd37ec3b5fd0b10dd#r1264642467) with the minio headers. Is that fine for now?

oauth/src/main/java/ch/cyberduck/core/oauth/OAuth2AuthorizationService.java
```diff
@@ -345,13 +354,14 @@ else if(value instanceof Number) {
...
-            return new TokenResponse()
+        public IdTokenResponse toTokenResponse() {
+            return new IdTokenResponse()
                     .setTokenType(tokenType)
                     .setScope(scope)
                     .setExpiresInSeconds(expiresInSeconds)
                     .setAccessToken(accessToken)
-                    .setRefreshToken(refreshToken);
+                    .setRefreshToken(refreshToken)
+                    .setIdToken(idToken);
```
dkocher, line 364: This causes a `NullPointerException` when no id token is returned.

SheromS: Good point. Is resolved by the latest commit.

core/src/main/java/ch/cyberduck/core/DefaultHostPasswordStore.java
```diff
@@ -272,6 +279,10 @@ public void delete(final Host bookmark) throws LocalAccessDeniedException {
             if(credentials.getOauth().getExpiryInMilliseconds() != null) {
                 this.deletePassword(this.getOAuthHostname(bookmark), String.format("%s OAuth2 Token Expiry", prefix));
             }
+            if(StringUtils.isNotBlank(credentials.getOauth().getRefreshToken())) {
```
dkocher, line 282: Check should be on `getIdToken`

.github/workflows/ci.yml
```diff
@@ -31,3 +31,12 @@ jobs:
       - name: Publish Test Report
         if: ${{ always() }}
         uses: ScaCap/action-surefire-report@v1.7.0
+
```
dkocher, line 34: Indent seems to be wrong making the file syntax invalid.

core/src/main/java/ch/cyberduck/core/DefaultHostPasswordStore.java
```diff
@@ -232,6 +234,11 @@ public void save(final Host bookmark) throws LocalAccessDeniedException {
                 this.addPassword(this.getOAuthHostname(bookmark), String.format("%s OAuth2 Token Expiry", prefix),
                         String.valueOf(credentials.getOauth().getExpiryInMilliseconds()));
             }
+            if(StringUtils.isNotBlank(credentials.getOauth().getIdToken())) {
+                this.addPassword(bookmark.getProtocol().getScheme(),
+                        bookmark.getPort(), this.getOAuthHostname(bookmark),
+                        String.format("%s OIDC Id Token", prefix), credentials.getOauth().getRefreshToken());
```
dkocher, line 240: Saving refresh token instead of id token.

oauth/src/main/java/ch/cyberduck/core/oauth/OAuth2AuthorizationService.java
```diff
@@ -135,25 +135,30 @@ public OAuthTokens authorize(final Host bookmark, final LoginCallback prompt, fi
...
         }
-        final TokenResponse response;
+
+        final IdTokenResponse response;
+        // Save access token, refresh token and id token
         switch(flowType) {
             case AuthorizationCode:
                 response = this.authorizeWithCode(bookmark, prompt);
-                break;
+                return credentials.withOauth(new OAuthTokens(
```
dkocher, line 144: It has been discussed before that these two return statement are identical and can be moved below the switch.

s3/src/test/java/ch/cyberduck/core/sts/OidcAuthenticationTest.java
```diff
@@ -0,0 +1,168 @@
...
+
+        // assert that STS credentials expires with the oAuth token even though the duration seconds is valid for longer
+        STSCredentialsRequestInterceptor authorizationService = session.getAuthorizationService();
+        assertTrue(40 > ((authorizationService.getStsExpiryInMilliseconds() - System.currentTimeMillis()) / 1000));
+        assertEquals(Optional.of(authorizationService.getStsExpiryInMilliseconds()).get() / 1000, authorizationService.getTokens().getExpiryInMilliseconds() / 1000);
+
+        session.close();
+    }
+
+    /** only use with the below specified changes in the keycloak config json file and run as separate test
```
dkocher, line 120: Ideally we find a way to run these tests with different configurations in `keycloak-realm.json`.

IsmCadi: My first thought was to use a second realm with these configurations, but that would lead to a second IDP config in MinIO which is not possible with the same address. As of now I do not know how to do this without maybe using additional containers, which would complicate the Testcontainers and the Docker config.

chenkins: @dkocher I have drafted a way to run test with different configurations in `keycloak-realm.json`.
`testSTSCredentialsExpiredValidOAuthToken` will take 15 minutes as the min STS duration is 900s see https://min.io/docs/minio/linux/developers/security-token-service/AssumeRoleWithWebIdentity.html. So unsure what it makes sense to finish it.

@IsmCadi I struggle to see the  working of `testBucketRequestBeforeTokenExpiryFailsBecauseOfLatency` - can you re-phrase what you're testing and how to adjust the 28 secs? The test is red when I run locally.

chenkins: Let's continue discussion here: https://github.com/iterate-ch/cyberduck/pull/14972/files#r1291199162

chenkins:
Closing this pr after discussion with @dkocher -  superseded by #14972 .

PR #14907, opened 2023-07-17
Modified timestamp in transfer status needs to be set for existing files too.

ylangisc:


googledrive/src/main/java/ch/cyberduck/core/googledrive/DriveWriteFeature.java
```diff
@@ -120,8 +121,13 @@ public File call(final AbstractHttpEntity entity) throws BackgroundException {
                         switch(postResponse.getStatusLine().getStatusCode()) {
                             case HttpStatus.SC_OK:
                                 if(status.isExists()) {
-                                    return session.getClient().getObjectParser().parseAndClose(
+                                    final File f = session.getClient().getObjectParser().parseAndClose(
                                             new InputStreamReader(postResponse.getEntity().getContent(), StandardCharsets.UTF_8), File.class);
+                                    if(null != status.getTimestamp()) {
+                                        new DriveTimestampFeature(session, fileid).setTimestamp(file, status.getTimestamp());
```
dkocher, line 127: I would duplicate the implementation which allows to directly return the updated `File` reference.

```suggestion
            final File properties = new File();
            properties.setModifiedTime(new DateTime(status.getTimestamp()));
            return session.getClient().files().update(fileid, properties).setFields(DriveAttributesFinderFeature.DEFAULT_FIELDS).
                    setSupportsAllDrives(new HostPreferences(session.getHost()).getBoolean("googledrive.teamdrive.enable")).execute();
```

dkocher (changes requested):

googledrive/src/main/java/ch/cyberduck/core/googledrive/DriveWriteFeature.java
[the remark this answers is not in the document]
dkocher: In general we do not handle requests gracefully but allow to retry the transfer.

AliveDevil: Create implementation method in Timestamp-feature that can be used from both Timestamp-feature and Write-feature, without duplicating code.

ylangisc: The `update` method must not contain fields that are not modifiable. The suggested change thus leads to 403 `The resource body includes fields which are not directly writable`.

AliveDevil (approved):

googledrive/src/main/java/ch/cyberduck/core/googledrive/DriveWriteFeature.java
dkocher, line 127:
```suggestion
                                        new DriveTimestampFeature(session, fileid).setTimestamp(file, status);
```

PR #14920, opened 2023-07-21
Clear error state for the case a retry succeeded.

ylangisc:


dkocher:
Override `retry` and reset canceled state there

```
                        @Override
                        public boolean retry(final BackgroundException failure, final ProgressListener progress, final BackgroundActionState cancel) {
                            if(super.retry(failure, progress, cancel)){
                                canceled.set(null);
                                return true;
                            }
                            return false;
                        }
```

dkocher:
Additionally apply same fix to other implementations using same pattern.
- `BrickMultipartWriteFeature`
- `SDSDirectS3MultipartWriteFeature`
- `EueMultipartWriteFeature`
- `S3MultipartWriteFeature`
- `StoregateMultipartWriteFeature`

dkocher (approved):

Issue #14590, opened 2023-04-28
Wrong processing of folder/placeholder names starting with a ~


ylangisc:
Listing a folder with a leading `~` duplicates itself in the result set, e.g.

Expected result for listing `/bla/~myfolder`
```
file1.txt
file2.txt
...
```

Current, wrong result
```
~myfolder
file1.txt
file2.txt
```

PR #14951, opened 2023-07-31
Allow tilde to denote home directory in default path. 

### LINKED ISSUE #14936: Wrong directory opened when choosing open in Terminal
**Describe the bug**
I have an open ssh connection in Cyberduck and would like to directly open an active ssh terminal connection. Therefore I press the "terminal" button in the window. But instead I get an error or an open ssh connection to the wrong directory.

## Scenario 1:
Press the "terminal" button. Then the terminal window opens with this command:
```sh
ssh -t -i "/Users/<username>/.ssh/id_rsa" <user>@<host> -p 22 "cd \/ && exec \$SHELL -l"
```
The problematic part is ```cd \/```, because this switches into the root folder, although I would like the folder not to change at all. But at least the connection gets up and running so I can ```cd ~``` back manually into the right folder. 

Which is not the case, when I have configured a relative (!) folder path in Cyberduck for this connection.

## Scenario 2:
When I have a relative Path configured (in this example ```relativePath```), and I press the "terminal" button this SSH command gets send to the terminal:
```sh
~ » ssh -t -i "/Users/< username >/.ssh/id_rsa" <user>@<host> -p 22 "cd \/relativePath && exec \$SHELL -l"
```
Right after starting the connecting it tries to switch into a root folder called "relativePath". But "relativePath" is not located in root. Therefore the connection can not be be built up and an error occurs. In my case the error (in German) is:
```
bash: Zeile 0: cd: /logs: Datei oder Verzeichnis nicht gefunden
Connection to <host> closed.
```

**To Reproduce**
1. Configure SSH Connection with a path configured and save it
2. Open the connection
3. Open the terminal connection in the window

**Expected behavior**
I expect Cyberduck to open the terminal connection right at the same spot, where my Cyberduck connection is configured (or even better: at the same folder, I am browsing at the moment)

 - OS: Ventura 13.4.1, M1
 - Version 8.6.0

**Additional context**
I have set up my Cyberduck connection with an id.rsa key for authorization. I don't know if this changes anything.

dkocher:
Fix #14936.

### DIFF
--- a/core/src/main/java/ch/cyberduck/core/shared/DefaultPathHomeFeature.java
+++ b/core/src/main/java/ch/cyberduck/core/shared/DefaultPathHomeFeature.java
@@ -18,31 +18,44 @@ package ch.cyberduck.core.shared;
  * feedback@cyberduck.ch
  */
 
-import ch.cyberduck.core.Host;
 import ch.cyberduck.core.Path;
 import ch.cyberduck.core.PathNormalizer;
+import ch.cyberduck.core.Session;
+import ch.cyberduck.core.TildePathExpander;
 import ch.cyberduck.core.exception.BackgroundException;
+import ch.cyberduck.core.features.Home;
 
 import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
+import java.util.EnumSet;
+
 public class DefaultPathHomeFeature extends AbstractHomeFeature {
     private static final Logger log = LogManager.getLogger(DefaultPathHomeFeature.class);
 
-    private final Host host;
+    private final Session<?> session;
 
-    public DefaultPathHomeFeature(final Host host) {
-        this.host = host;
+    public DefaultPathHomeFeature(final Session<?> session) {
+        this.session = session;
     }
 
     @Override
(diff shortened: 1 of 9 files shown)


core/src/main/java/ch/cyberduck/core/shared/DefaultPathHomeFeature.java
```diff
@@ -18,31 +18,44 @@
...
+    public DefaultPathHomeFeature(final Session<?> session) {
+        this.session = session;
     }
 
     @Override
     public Path find() throws BackgroundException {
-        if(StringUtils.isNotBlank(host.getDefaultPath())) {
-            return PathNormalizer.compose(ROOT, host.getDefaultPath());
+        if(StringUtils.isNotBlank(session.getHost().getDefaultPath())) {
+            if(StringUtils.startsWith(session.getHost().getDefaultPath(), Path.HOME)) {
```
ylangisc, line 46: `~myfolder/subfolder` should not expand but `~/myfolder` and `~`.

dkocher: Relates to #14590.

ylangisc (approved):

dkocher:
Addendum in #14959.

PR #14972, opened 2023-08-07
Allow connection profiles to use custom STS and OIDC endpoints

### LINKED ISSUE #13804: Suport to login using temporary credentials from Security Token Service (STS API) using OpenID Connect web identity

dkocher:
Resolves #13804.


oauth/src/main/java/ch/cyberduck/core/oauth/OAuth2AuthorizationService.java
```diff
@@ -135,25 +135,30 @@ public OAuthTokens authorize(final Host bookmark, final LoginCallback prompt, fi
...
         }
-        final TokenResponse response;
+
+        final IdTokenResponse response;
+        // Save access token, refresh token and id token
         switch(flowType) {
             case AuthorizationCode:
                 response = this.authorizeWithCode(bookmark, prompt);
-                break;
+                return credentials.withOauth(new OAuthTokens(
```
chenkins, line 144: Duplicated code - refactor into method?

s3/src/main/java/ch/cyberduck/core/sts/STSAssumeRoleTokenExpiredResponseInterceptor.java
```diff
@@ -0,0 +1,109 @@
...
+                // Break through
+            case HttpStatus.SC_UNAUTHORIZED:
+                if(executionCount <= MAX_RETRIES) {
+                    try {
+                        log.warn(String.format("Attempt to refresh STS token for failure %s", response));
+                        final STSTokens tokens = sts.refresh(oauth.getTokens());
+                        session.getClient().setProviderCredentials(new AWSSessionCredentials(tokens.getAccessKey(),
+                                tokens.getSecretAccessKey(), tokens.getSessionToken()));
+                        // Try again
+                        return true;
```
chenkins, line 96: I'd need a flow/state diagram 😉

s3/src/main/java/ch/cyberduck/core/s3/S3Session.java
```diff
@@ -201,12 +206,29 @@ protected String getRestMetadataPrefix() {
...
-                    new ThreadLocalHostnameDelegatingTrustManager(trust, host.getHostname()), key, prompt));
+        if(host.getProtocol().isOAuthConfigurable()) {
+            configuration.addInterceptorLast(oauth = new OAuth2RequestInterceptor(builder.build(ProxyFactory.get()
+                    .find(host.getProtocol().getOAuthAuthorizationUrl()), this, prompt).build(), host)
+                    .withRedirectUri(host.getProtocol().getOAuthRedirectUrl())
+                    .withFlowType(OAuth2AuthorizationService.FlowType.valueOf(host.getProtocol().getAuthorization())));
+            configuration.addInterceptorLast(sts = new STSAssumeRoleCredentialsRequestInterceptor(oauth, this, trust, key));
+            configuration.setServiceUnavailableRetryStrategy(new STSAssumeRoleTokenExpiredResponseInterceptor(this, oauth, sts, prompt));
+        }
+        else {
```
chenkins, line 219: contract to `else if` for readability?

s3/src/main/java/ch/cyberduck/core/s3/S3Session.java
```diff
@@ -201,12 +206,29 @@ protected String getRestMetadataPrefix() {
...
-        // Only for AWS
-        if(S3Session.isAwsHostname(host.getHostname())) {
-            configuration.setServiceUnavailableRetryStrategy(new S3TokenExpiredResponseInterceptor(this,
-                    new ThreadLocalHostnameDelegatingTrustManager(trust, host.getHostname()), key, prompt));
+        if(host.getProtocol().isOAuthConfigurable()) {
+            configuration.addInterceptorLast(oauth = new OAuth2RequestInterceptor(builder.build(ProxyFactory.get()
+                    .find(host.getProtocol().getOAuthAuthorizationUrl()), this, prompt).build(), host)
+                    .withRedirectUri(host.getProtocol().getOAuthRedirectUrl())
+                    .withFlowType(OAuth2AuthorizationService.FlowType.valueOf(host.getProtocol().getAuthorization())));
+            configuration.addInterceptorLast(sts = new STSAssumeRoleCredentialsRequestInterceptor(oauth, this, trust, key));
```
chenkins, line 216: As `sts` comes last, does this mean we renew `oauth` even if STS token is still valid? Is this necessary? Comment on this. Do we have tests for all possible cases?

s3/src/main/java/ch/cyberduck/core/s3/S3Session.java
```diff
@@ -201,12 +206,29 @@ protected String getRestMetadataPrefix() {
...
+            configuration.setServiceUnavailableRetryStrategy(new STSAssumeRoleTokenExpiredResponseInterceptor(this, oauth, sts, prompt));
+        }
+        else {
+            if(S3Session.isAwsHostname(host.getHostname())) {
+                // Try auto-configure
```
chenkins, line 221: As auto-configure part is moved to the `else` part, does the behaviour of existing configurations change? Danger of regressions? Documentation?

s3/src/main/java/ch/cyberduck/core/s3/S3TokenExpiredResponseInterceptor.java
```diff
@@ -41,24 +40,24 @@ public class S3TokenExpiredResponseInterceptor extends DisabledServiceUnavailabl
...
             switch(response.getStatusLine().getStatusCode()) {
                 case HttpStatus.SC_BAD_REQUEST:
-                    final S3ServiceException failure;
                     try {
                         if(null != response.getEntity()) {
                             EntityUtils.updateEntity(response, new BufferedHttpEntity(response.getEntity()));
-                            failure = new S3ServiceException(response.getStatusLine().getReasonPhrase(),
+                            final S3ServiceException failure = new S3ServiceException(response.getStatusLine().getReasonPhrase(),
                                     EntityUtils.toString(response.getEntity()));
+                            failure.setResponseCode(response.getStatusLine().getStatusCode());
```
chenkins, line 60: Why is this necessary now and was not necessary before?

s3/src/main/java/ch/cyberduck/core/sts/STSAssumeRoleCredentialsRequestInterceptor.java
```diff
@@ -0,0 +1,106 @@
...
+        if(log.isDebugEnabled()) {
+            log.debug(String.format("Save new tokens %s for %s", tokens, host));
+        }
+        store.save(host);
+        return tokens;
+    }
+
+    @Override
+    public void process(final HttpRequest request, final HttpContext context) throws HttpException, IOException {
+        if(tokens.isExpired()) {
```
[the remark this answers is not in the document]
dkocher: There is no local variable `tokens` in this method context.

s3/src/main/java/ch/cyberduck/core/s3/S3TokenExpiredResponseInterceptor.java
dkocher: This was a bug in the existing code base filed as #10917.

s3/src/main/java/ch/cyberduck/core/sts/STSAssumeRoleTokenExpiredResponseInterceptor.java
dkocher: Credentials are accessed from `client.getProviderCredentials()` in request interceptors and thus available upon retry of the request.

s3/src/main/java/ch/cyberduck/core/s3/S3Session.java
dkocher: Which part should be expanded?

core/src/main/java/ch/cyberduck/core/KeychainLoginService.java
```diff
@@ -158,6 +159,9 @@ public boolean prompt(final Host bookmark, final String message, final LoginCall
             credentials.setToken(input.getPassword());
         }
         if(options.oauth) {
+            prompt.warn(bookmark, LocaleFactory.localizedString("Login failed", "Credentials"), message,
+                    LocaleFactory.localizedString("Continue", "Credentials"),
```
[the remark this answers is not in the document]
dkocher: Will change to _Try Again_.

oauth/src/main/java/ch/cyberduck/core/oauth/OAuth2AuthorizationService.java
dkocher: Can just be moved after the switch statement.

s3/src/main/java/ch/cyberduck/core/sts/STSAssumeRoleTokenExpiredResponseInterceptor.java
```diff
@@ -0,0 +1,109 @@
...
+                        else {
+                            // Ignore other 400 failures
+                            return false;
+                        }
+                    }
+                }
+                catch(IOException e) {
+                    log.warn(String.format("Failure parsing response entity from %s", response));
+                }
+                // Break through
```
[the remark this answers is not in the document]
dkocher: Refactored in 1ef84c15.

s3/src/test/resources/S3-OIDC-Testing.cyberduckprofile
```diff
@@ -0,0 +1,65 @@
...
+        <key>Password Configurable</key>
+        <true/>
+        <key>Username Configurable</key>
+        <true/>
+        <key>Path Configurable</key>
+        <true/>
+        <key>Properties</key>
+        <array>
+            <string>s3.bucket.virtualhost.disable=true</string>
+            <string>s3.assumerole.rolearn=</string>
```
chenkins, line 62: necessary?

s3/src/test/resources/testcontainer/docker-compose.yml
```diff
@@ -0,0 +1,89 @@
...
+
+  keycloak:
+    hostname: keycloak
+    build: keycloak
+    ports:
+      - "8080:8080"
+    environment:
+      KEYCLOAK_ADMIN: admin
+      KEYCLOAK_ADMIN_PASSWORD: admin
+      KEYCLOAK_IMPORT: /tmp/keycloak-realm.json
```
chenkins, line 13: Obsolete, we mount volumes below.

s3/src/main/java/ch/cyberduck/core/sts/STSAssumeRoleAuthorizationService.java
```diff
@@ -0,0 +1,163 @@
...
+        final AssumeRoleWithWebIdentityRequest request = new AssumeRoleWithWebIdentityRequest();
+        final String token;
+        if(StringUtils.isNotBlank(oauth.getIdToken())) {
+            if(log.isDebugEnabled()) {
+                log.debug(String.format("Assume role with OIDC Id token for %s", bookmark));
+            }
+            token = oauth.getIdToken();
+        }
+        else {
+            if(log.isDebugEnabled()) {
```
dkocher, line 109: @chenkins Should we skip this and require an OIDC Id token ?

s3/src/main/java/ch/cyberduck/core/sts/STSAssumeRoleAuthorizationService.java
```diff
@@ -0,0 +1,163 @@
...
+                if(StringUtils.isNotBlank(sub)) {
+                    request.setRoleSessionName(sub);
+                }
+                else {
+                    log.warn(String.format("Missing subject in decoding JWT %s", token));
+                    request.setRoleSessionName(new AsciiRandomStringService().random());
+                }
+            }
+            catch(JWTDecodeException e) {
+                log.warn(String.format("Failure decoding JWT %s", token));
```
dkocher, line 140: @chenkins We should probably fail fast here with an exception if the token cannot be decoded as a JWT?

s3/src/main/java/ch/cyberduck/core/sts/STSAssumeRoleAuthorizationService.java
dkocher: In 204949f8

s3/src/main/java/ch/cyberduck/core/sts/STSAssumeRoleAuthorizationService.java
dkocher: In 204949f8

s3/src/test/java/ch/cyberduck/core/sts/OidcAuthenticationTest.java
```diff
@@ -0,0 +1,154 @@
...
+        session.close();
+    }
+
+    /** only use with the below specified changes in the keycloak config json file and run as separate test
+     * set config keycloak-realm.json:
+     *      "access.token.lifespan": "930"
+     *      "ssoSessionMaxLifespan": 1100,
+     */
+    @Test
+    @Ignore
```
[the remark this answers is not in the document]
chenkins: Moving discussion here from https://github.com/iterate-ch/cyberduck/pull/14893#discussion_r1289784236 

 @dkocher:
> Ideally we find a way to run these tests with different configurations in `keycloak-realm.json`.


@IsmCad:
> i My first thought was to use a second realm with these configurations, but that would lead to a second IDP config in MinIO which is not possible with the same address. As of now I do not know how to do this without maybe using additional containers, which would complicate the Testcontainers and the Docker config.


@chenkins 
> @dkocher I have drafted a way to run test with different configurations in `keycloak-realm.json`. `testSTSCredentialsExpiredValidOAuthToken` will take 15 minutes as the min STS duration is 900s see https://min.io/docs/minio/linux/developers/security-token-service/AssumeRoleWithWebIdentity.html. So unsure what it makes sense to finish it.
> 
> @IsmCadi I struggle to see the working of `testBucketRequestBeforeTokenExpiryFailsBecauseOfLatency` - can you re-phrase what you're testing and how to adjust the 28 secs? The test is red when I run locally.

s3/src/main/java/ch/cyberduck/core/s3/S3ExceptionMappingService.java
```diff
@@ -73,6 +73,7 @@ public BackgroundException map(final ServiceException e) {
                             return new ConnectionTimeoutException(buffer.toString(), e);
                         case "ExpiredToken":
                         case "InvalidToken":
+                        case "TokenRefreshRequired":
```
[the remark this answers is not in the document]
dkocher: Can be found at https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html#ErrorCodeList

s3/src/main/java/ch/cyberduck/core/s3/S3Session.java
dkocher: Implementation currently is intended as follows
1. Refresh of OAuth tokens when expired preemptively in `OAuth2RequestInterceptor` [^1] 
2. Refresh STS tokens using cached OAuth token preemptively in `STSAssumeRoleCredentialsRequestInterceptor` [^2]
3. On error response handle cases in `STSAssumeRoleTokenExpiredResponseInterceptor` as follows

* On `401`, `403` refresh OIDC Id token
* On `400` reuse cached OIDC Id token
* For both cases above refresh STS tokens

[^1]: https://github.com/iterate-ch/cyberduck/pull/14972/files#diff-7ec37d50b17b31a07e5438bce7f0f3f4eba8a930b1c100e12e3ad84dda666ff8R106
[^2]: https://github.com/iterate-ch/cyberduck/pull/14972/files#diff-c51064e4ed9ed159481473d881ec27aae4bd319978568b45480976f8f989b043R89

s3/src/main/java/ch/cyberduck/core/s3/S3Session.java
```diff
@@ -269,162 +291,40 @@ public void process(final HttpRequest request, final HttpContext context) {
...
+                    tokens.getSecretAccessKey(), tokens.getSessionToken()));
         }
         else {
             final Credentials credentials;
             // Only for AWS
             if(isAwsHostname(host.getHostname())) {
-                // Try auto-configure
-                credentials = new AWSProfileSTSCredentialsConfigurator(
-                        new ThreadLocalHostnameDelegatingTrustManager(trust, host.getHostname()), key, prompt).configure(host);
+                try {
```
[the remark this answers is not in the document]
dkocher: Would defer this to a later refactoring.

s3/src/main/java/ch/cyberduck/core/s3/S3Session.java
```diff
@@ -269,162 +291,40 @@ public void process(final HttpRequest request, final HttpContext context) {
...
+                    credentials = token.refresh();
+                }
+                catch(ConnectionTimeoutException | ConnectionRefusedException | ResolveFailedException |
+                      NotfoundException | InteroperabilityException e) {
+                    log.warn(String.format("Failure to retrieve session credentials from . %s", e.getMessage()));
+                    throw new LoginFailureException(e.getDetail(false), e);
+                }
             }
             else {
                 credentials = host.getCredentials();
```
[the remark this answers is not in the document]
dkocher: The credentials configurator will return the same credentials when no change is required.

s3/src/test/java/ch/cyberduck/core/sts/OidcAuthenticationTest.java
```diff
@@ -0,0 +1,154 @@
...
+        session.open(Proxy.DIRECT, new DisabledHostKeyCallback(), new DisabledLoginCallback(), new DisabledCancelCallback());
+        session.login(Proxy.DIRECT, new DisabledLoginCallback(), new DisabledCancelCallback());
+
+        String firstAccessToken = host.getCredentials().getOauth().getIdToken();
+        String firstAccessKey = session.getClient().getProviderCredentials().getAccessKey();
+        String firstSessionToken = ((AWSSessionCredentials) session.getClient().getProviderCredentials()).getSessionToken();
+
+        Path container = new Path("cyberduckbucket", EnumSet.of(Path.Type.directory, Path.Type.volume));
+        assertTrue(new S3FindFeature(session, new S3AccessControlListFeature(session)).find(container));
+        Thread.sleep(1100 * 30);
```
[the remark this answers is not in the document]
IsmCadi: Yes, thats right.

s3/src/test/java/ch/cyberduck/core/sts/OidcAuthenticationTest.java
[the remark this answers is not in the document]
SheromS: If an HTTP HEAD request to MinIO fails, the error code and error description are written to MinIO-specific error headers, since HTTP HEAD requests do not use a body. For example, a HEAD request is used to check whether an S3 bucket or folder is discoverable.
This issue came to light when a HEAD request uses expired STS credentials. Because of preemtively refresh tokens this case is only possible it the credentials are still valid but a few milliseconds before expire. Because of the latency in the network the request will be invalid when reaching the MinIO Service. But the sleep time needs to be ajusted according to the network latency.
Overall the test may be removed and the general question is how to handle the MinIO-specific HTTP-Headers when a HEAD-Request is failing.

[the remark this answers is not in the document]
dkocher: We are not interested in the special case of missing response body in HEAD requests where we have the S3 error code in the response header instead (handled in 56b96842) but rather if we can differentiate from the S3 API or STS API error if the OpenID Id token must be refreshed or not if missed in the preemptive handler.

s3/src/main/java/ch/cyberduck/core/s3/S3Session.java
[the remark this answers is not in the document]
dkocher: In 7b22c8a5.

s3/src/main/java/ch/cyberduck/core/s3/S3Session.java
dkocher: Updated handling of authentication failures in `S3AuthenticationResponseInterceptor` as of 8ea744a1.

* Handle `403` and `400`
* Refresh STS token with cached OIDC Id token when not expired
* Otherwise refresh OIDC Id token first
* On failure refresh STS token with refreshed OIDC Id token

| Google OpenID Provider  | AWS API                       | HTTP Status Code | Error Code (in XML response body) |
|-------------------------|-------------------------------|------------------|-----------------------------------|
| OpenID Id Token Expired | S3 ListObjects                |                  | [^1]                              |
| OpenID Id Token Expired | STS AssumeRoleWithWebIdentity | 400              | `ExpiredTokenException` [^2]      |
| STS Tokens Expired      | S3 ListObjects                | 400              | `ExpiredToken` [^3]               |
| STS Tokens Invalid      | S3 ListObjects                | 400              | `InvalidToken` [^4]               |

[^1]:
[^2]: `Token expired: current date/time 1691790439 must be before the expiration date/time1691788167`

```
HTTP/1.1 400 Bad Request
x-amzn-RequestId: b6baa352-afcb-4763-a32c-a67d7eb5020b
Content-Type: text/xml
Content-Length: 346
Date: Sun, 13 Aug 2023 10:53:10 GMT
Connection: close

<ErrorResponse xmlns="https://sts.amazonaws.com/doc/2011-06-15/">
  <Error>
    <Type>Sender</Type>
    <Code>ExpiredTokenException</Code>
    <Message>Token expired: current date/time 1691923690 must be before the expiration date/time1691878565</Message>
  </Error>
  <RequestId>b6baa352-afcb-4763-a32c-a67d7eb5020b</RequestId>
</ErrorResponse>
```

[^3]: `The provided token has expired.`
[^3]: `The provided token is malformed or otherwise invalid.`

core/src/main/java/ch/cyberduck/core/Credentials.java
```diff
@@ -329,23 +347,25 @@ public boolean equals(final Object o) {
         final Credentials that = (Credentials) o;
         return Objects.equals(user, that.user) &&
             Objects.equals(password, that.password) &&
-            Objects.equals(token, that.token) &&
+                Objects.equals(token, that.token) &&
```
ylangisc, line 350: Formatting.

core/src/main/java/ch/cyberduck/core/Credentials.java
```diff
@@ -73,6 +74,7 @@ public Credentials(final Credentials copy) {
         this.user = copy.user;
         this.password = copy.password;
         this.token = copy.token;
+        this.tokens = copy.tokens;
```
ylangisc, line 77: Confusing naming as we already have `token`.

core/src/main/java/ch/cyberduck/core/Credentials.java
```diff
@@ -329,23 +347,25 @@ public boolean equals(final Object o) {
...
+                Objects.equals(token, that.token) &&
+            Objects.equals(tokens, that.tokens) &&
             Objects.equals(identity, that.identity) &&
             Objects.equals(certificate, that.certificate);
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(user, password, token, identity, certificate);
+        return Objects.hash(user, password, tokens, identity, certificate);
```
ylangisc, line 358: `equals` honours `token` _and_ `tokens`. Should be symmetric.

core/src/main/java/ch/cyberduck/core/Profile.java
```diff
@@ -525,7 +525,7 @@ public String getOAuthClientId() {
     @Override
     public String getOAuthClientSecret() {
         final String v = this.value(OAUTH_CLIENT_SECRET_KEY);
-        if(StringUtils.isBlank(v)) {
+        if(null == v) {
```
ylangisc, line 528: Why check for `null`? All other fields are checked against `StringUtils.isBlank()`.

dkocher: Because this allow to define a default empty client secret that skip the prompt. See cbb5a225 and 43dc41b3.

ylangisc (changes requested):
https://github.com/iterate-ch/cyberduck/pull/14972/files#r1299842670

ylangisc (approved):

PR #14978, opened 2023-08-09
Allow substitution of profile values and properties using `${}` with …

### LINKED ISSUE #14977: Allow substitution of variables in connection profiles with preferences
Allow values in connection profiles to contain `${…}` and replace when reading with value from preferences.

dkocher:
…value from preferences. Resolves #14977.

### DIFF
--- a/core/src/main/java/ch/cyberduck/core/Profile.java
+++ b/core/src/main/java/ch/cyberduck/core/Profile.java
@@ -29,11 +29,14 @@ import ch.cyberduck.core.serializer.Serializer;
 import org.apache.commons.codec.binary.Base64;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.text.StringSubstitutor;
+import org.apache.commons.text.lookup.StringLookupFactory;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
 import java.io.IOException;
 import java.io.OutputStream;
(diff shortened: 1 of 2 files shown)


core/src/main/java/ch/cyberduck/core/Profile.java
```diff
@@ -570,15 +576,17 @@ public boolean isUTCTimezone() {
...
         if(null == list) {
             return Collections.emptyList();
         }
-        return list;
+        final ArrayList<String> substituted = new ArrayList<>(list);
+        substituted.replaceAll(substitutor::replace);
+        return substituted;
     }
 
     private boolean bool(final String key) {
```
chenkins, line 592: no support for injecting `bool`s?

core/src/main/java/ch/cyberduck/core/Profile.java
```diff
@@ -29,11 +29,14 @@
 import org.apache.commons.codec.binary.Base64;
```
chenkins, line 29: No unit tests in this mr?

chenkins, line 29: No unit tests in this mr?

core/src/main/java/ch/cyberduck/core/Profile.java
dkocher: Not possible as the property list dtd does not allow arbitrary values in this case.

AliveDevil (approved):

PR #14988, opened 2023-08-11
Slimmed keycloak-realm.json.

chenkins:


dkocher (approved):
That makes it a lot more readable 👍🏻

PR #14991, opened 2023-08-12
Test setup OIDC inject keycloak

chenkins:


dkocher:
Cherry-picked in 17a4588e.

PR #14993, opened 2023-08-13
Fix #13801.

### LINKED ISSUE #13801: Opening two browser window after launching application
Cyberduck opens a second panel while starting. 

Reported for the standalone version on macOS

dkocher:
Resolves #13801.


ylangisc (commented):
Should be merged with default bookmark handling below.

ylangisc (approved):

PR #15009, opened 2023-08-18
SMB for Cyberduck

### LINKED ISSUE #5368: Support for Samba (SMB) protocol
> **841fd6e** created the issue

Is there a possibility to add ability to access SMB (Samba) servers to CyberDuck?

The Finder implementation is spotty.

mbscit:


dkocher:
Can you clarify for me the following points to allow us to open issues for pending tasks:

- [ ] Was DFS tested with the option `dfsEnabled` set in `SmbConfig`? The latest version of smbj includes changes [^1]  that possibly resolves issues. This could be tested by updating the dependency to `0.12.2`
- [ ] You introduced `SmbPath` to translate path delimiters. It looks like this is already handled in the library [^2]. Can you pinpoint where this failed prior introducing the custom converter for path strings? I see no test failure when omitting.
- [x] Authenticating issues connecting to macOS [^2]  can no longer be reproduced with `0.12.2`. 

[^1]: hierynomus/smbj#722)
[^2]: https://github.com/hierynomus/smbj/blob/d6b2fadaea2df96111a316dab75a14dc7c4142d9/src/main/java/com/hierynomus/smbj/common/SmbPath.java
[^3]: hierynomus/smbj#730

mbscit:
1. `dfsEnabled` was tested before, but not with 0.12.2
Just tested it, works better now but still has some problems:

- [ ]  It often runs into timeouts, especially on the list feature.
- [ ] Move / rename and edit do not work (STATUS_PRIVILEGE_NOT_HELD) but the same file/folder can be renamed and edited using macOS Finder

2. The path delimiters have caused problems with the move feature on Windows shares (STATUS_OBJECT_NAME_INVALID), this was the reason for the conversion.
Just retested it with 0.12.2, still the same issue.


3. We can confirm that login now works with macOS shares. However, most features did not work properly in our tests:
- The File size was always 0 Byte (List, Upload, Download, Edit etc.)
- STATUS_SHARING_VIOLATION on Rename/Move

But we also had some problems with the share using macOS Finder, the share was behaving strangely. We recommend testing with another share.


We have updated the PR with smbj 0.12.2

dkocher (approved):

PR #15033, opened 2023-08-24
Fix #15031

### LINKED ISSUE #15031: OAuth access token repeatedly saved in keychain
OAuth tokens obtained with no refresh token are repeatedly saved in keychain with no change.

dkocher:


oauth/src/main/java/ch/cyberduck/core/oauth/OAuth2RequestInterceptor.java
```diff
@@ -105,7 +103,12 @@ public OAuthTokens save(final OAuthTokens tokens) throws LocalAccessDeniedExcept
     public void process(final HttpRequest request, final HttpContext context) throws HttpException, IOException {
         if(tokens.isExpired()) {
             try {
-                this.save(this.refresh(tokens));
+                final OAuthTokens previous = tokens;
+                final OAuthTokens refreshed = this.refresh(tokens);
+                // Skip saving tokens when not changed
+                if(!refreshed.equals(previous)) {
```
ylangisc, line 109: Implement `equals` for `OAuthTokens` would make it cleaner as we do not rely on the returning the same token object if there is no refresh token.

ylangisc (approved):

PR #15062, opened 2023-08-29
Fix #15061.

dkocher:


s3/src/main/java/ch/cyberduck/core/s3/S3Session.java
```diff
@@ -266,7 +266,12 @@ public void process(final HttpRequest request, final HttpContext context) {
                 // Ensure the required Host header is set prior to signing.
                 final HttpHost host = (HttpHost) context.getAttribute(HttpCoreContext.HTTP_TARGET_HOST);
                 if(host != null) {
-                    request.setHeader(HttpHeaders.HOST, host.toHostString());
+                    if(host.getPort() == S3Session.this.host.getProtocol().getDefaultPort()) {
+                        request.setHeader(HttpHeaders.HOST, host.getHostName());
```
ylangisc, line 270: Having my own S3 profile with a custom default port for my MiniIO instance would invalidate all signatures.

ylangisc (approved):

s3/src/main/java/ch/cyberduck/core/s3/S3Session.java
dkocher: Updated to compare default port in scheme instead of protocol.

PR #15096, opened 2023-09-06
Introduce property to allow skipping certain files from decrypting.

ylangisc:


core/src/main/java/ch/cyberduck/core/vault/DecryptingListProgressListener.java
```diff
@@ -16,33 +16,50 @@
...
         this.session = session;
         this.vault = vault;
         this.delegate = delegate;
+        this.skip = skip;
     }
 
     @Override
     public void visit(final AttributedList<Path> list, final int index, final Path f) {
         try {
+            if(!skip.accept(f)) {
```
dkocher, line 56: Would probably be better to inverse the logic and skip on `true`. Thus make the filter return `true` when the skip pattern matches.

dkocher (changes requested):

dkocher (approved):

PR #15105, opened 2023-09-08
Crash Reporter using Microsoft App Center

dkocher:


dkocher:
No longer pursuing as App Center is scheduled for retirement [^1].

[^1]: https://learn.microsoft.com/en-us/appcenter/retirement

PR #15114, opened 2023-09-12
Make OAuth prefix used for storing credentials configurable; use port…

### LINKED ISSUE #15113: Make OAuth prefix used for storing credentials configurable; use port from token URL instead from protocol
The same OAuth tokens might be shared between different protocols. The current implementation triggers OAuth flow for each different protocol description. Furthermore, in the case of OAuth Token Url, the bookmark's port is used instead of the token Url's port.

### LINKED ISSUE #15284: Shared OAuth refresh token not accepted
From #15230. OAuth tokens saved in the credentials manager shared between _Cyberduck_ and _Mountain Duck_ seem to cause issues for OneDrive although using the same registered OAuth Client ID for _Cyberduck_ and _Mountain Duck_.

```
AADSTS70000: The provided value for the input parameter 'refresh_token' or 'assertion' is not valid. The token was issued for a different client id
```

chenkins:
Resolves #15113


core/src/main/java/ch/cyberduck/core/DefaultHostPasswordStore.java
```diff
@@ -215,29 +216,65 @@ public void save(final Host bookmark) throws LocalAccessDeniedException {
...
                         String.format("%s OIDC Id Token", prefix), credentials.getOauth().getIdToken());
             }
         }
     }
 
+    private static int getPortFromTokenUrl(final Host bookmark) {
+        final String oAuthTokenUrl = bookmark.getProtocol().getOAuthTokenUrl();
+        int port = bookmark.getPort();
+        try {
+            URL url = new URL(oAuthTokenUrl);
```
dkocher, line 247: Use `java.net.URI.create` because `java.net.URL` is prone to do a name lookup.

core/src/main/java/ch/cyberduck/core/DefaultHostPasswordStore.java
```diff
@@ -215,29 +216,65 @@ public void save(final Host bookmark) throws LocalAccessDeniedException {
...
+            if(url.getPort() != -1) {
+                port = url.getPort();
+            }
+            else {
+                port = url.getDefaultPort();
```
dkocher, line 252: Can use `Scheme.valueOf(url).getScheme())`

dkocher (changes requested):

AliveDevil:
On Windows the OAuth Token is attached to the Credential Manager entry for a host in an additional "page"/attribute.

Basically:
```
protocol://user@hostname:port/
|- username: user
|- password: …
\  Metadata
    |-  OAuth Access Token: …
    |- OAuth Refresh Token: …
    \  OAuth Expiry: …
```

Thus the configurable oauth prefix isn't used there.
To make this work an additional credential manager entry for the OAuth connection has to be created.

core/src/main/java/ch/cyberduck/core/DefaultHostPasswordStore.java
```diff
@@ -215,29 +217,56 @@ public void save(final Host bookmark) throws LocalAccessDeniedException {
...
+        catch(UnsupportedSchemeException e) {
+            if(log.isWarnEnabled()) {
+                log.warn(String.format("Could not parse  %s, using protocol port instead.", oAuthTokenUrl), e);
+            }
+        }
+        return port;
+    }
+
+    private static Scheme getSchemeFromTokenUrl(final Host bookmark) {
+        final String oAuthTokenUrl = bookmark.getProtocol().getOAuthTokenUrl();
```
chenkins, line 265: Is `oAuthTokenUrl` always defined here? Default to `bookmark.getProtocol().getScheme()` if not?

core/src/main/java/ch/cyberduck/core/OAuthPrefixService.java
```diff
@@ -0,0 +1,72 @@
...
+    String getHostname();
+
+    /**
+     * Retrieves an identifier, used to identify Credential Manager entries on
+     * Windows (i.e. onedrive, sharepoint, dropbox, etc.)
+     */
+    String getIdentifier();
+
+    /**
+     * Some PasswordStores don't need to store all ports, thus allow for nullable
```
chenkins, line 49: @AliveDevil required for backwards compatibility only?  Could you not build the URI w/ and w/o port as a fallback to fetch existing passwords, but store w/ port in the future in the Windows implementation? Then, we could get rid of `getNonDefaultPort`?

core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
```diff
@@ -208,26 +219,78 @@ public override void save(Host bookmark)
...
+            string hostname = default;
+            string username = credentials.getUsername();
+
+            if (withOAuth && protocol.isOAuthConfigurable())
             {
-                pathBuilder.Append(":" + bookmark.getHostname());
-                if (protocol.isPortConfigurable() && !Equals(protocol.getDefaultPort(), bookmark.getPort()))
+                isOAuth = true;
+                OAuthPrefixService oAuthPrefix = new OAuthPrefixServiceFactory().create(bookmark);
+                hostname = oAuthPrefix.getIdentifier();
```
chenkins, line 239: @AliveDevil looks dodgy to assign `identifier` to `hostname`

core/src/main/java/ch/cyberduck/core/OAuthPrefixServiceFactory.java
```diff
@@ -0,0 +1,33 @@
...
+            if (null == constructor) {
+                log.warn(String.format("No matching constructor for parameter %s", Host.class));
+                // Call default constructor for disabled implementations
+                return clazz.getDeclaredConstructor().newInstance();
+            }
+            return constructor.newInstance(bookmark);
+        } catch (InstantiationException | InvocationTargetException | IllegalAccessException
+                | NoSuchMethodException e) {
+            log.error(String.format("Failure loading oauthprefixservice class %s. %s", clazz, e.getMessage()));
+            return new DefaultOAuthPrefixService(bookmark);
```
chenkins, line 30: @AliveDevil Do we want this hard-coded fallback?

core/src/main/java/ch/cyberduck/core/DefaultOAuthPrefixService.java
```diff
@@ -0,0 +1,80 @@
...
+    private final Protocol protocol;
+
+    public DefaultOAuthPrefixService(Host bookmark) {
+        this.bookmark = bookmark;
+        this.protocol = bookmark.getProtocol();
+    }
+
+    @Override
+    public String getDescription() {
+        if (StringUtils.isNotBlank(bookmark.getCredentials().getUsername())) {
```
chenkins, line 35: @AliveDevil we expose `getUsername()` in the same interface - have `getDescription()` return `bookmark.getProtocol().getDescription();` in the default implementation and  extract processing to `DefaultHostPasswordStore`
```
private String getOAuthPrefix(OAuthPrefixService oAuthPrefix) {
        if (StringUtils.isNotBlank(oAuthPrefix.getUsername())) {
            return String.format("%s (%s)", oAuthPrefix.getDescription(),
                    oAuthPrefix.getUsername());
        }
        return oAuthPrefix.getDescription()
}
```
?

core/src/main/java/ch/cyberduck/core/preferences/Preferences.java
```diff
@@ -555,6 +555,7 @@ protected void setFactories() {
         this.setDefault("factory.connectiontimeout.class", DefaultConnectionTimeout.class.getName());
         this.setDefault("factory.authorizationcodeprovider.class", "ch.cyberduck.core.oauth.BrowserOAuth2AuthorizationCodeProvider");
         this.setDefault("factory.s3.pathcontainerservice.class", "ch.cyberduck.core.s3.S3PathContainerService");
+        this.setDefault("factory.oauthprefixservice.class", OAuthPrefixService.class.getName());
```
chenkins, line 558: @AliveDevil should this not be `ch.cyberduck.core.DefaultOAuthPrefixService` instead of the interface?

AliveDevil: Good catch. Should be.

core/src/main/java/ch/cyberduck/core/OAuthPrefixServiceFactory.java
AliveDevil: Alternative is to create `DisabledOAuthPrefixService` (like all other implementations) and create that.
As there is no DisabledOAuthPrefixService … this class is used by default.

core/src/main/java/ch/cyberduck/core/OAuthPrefixService.java
AliveDevil: Possible, though `duck:hostname:443?user=` will then be created in the Credential Manager, instead of `duck:hostname?user=`.

core/src/main/java/ch/cyberduck/core/DefaultOAuthPrefixService.java
AliveDevil: Move `static getOAuthPrefix` back to DefaultHostPasswordStore then.

core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
AliveDevil: Well, that's what I called the field, could change that to `identifier`, which holds either `OAuthPrefix.getIdentifier()` or `bookmark.getHostname()` (in non-OAuth path).
Which is then built to `duck:identifier:port?user=` (OAuth) or `duck:fqdn:port?user=` (Default).

core/src/main/java/ch/cyberduck/core/OAuthPrefixService.java
dkocher: Always save port regardless of default.

core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
dkocher: This causes a regression as previously entries were saved using `bookmark.getProtocol().getOAuthTokenUrl()` when available.

dkocher: Only using the protocol identifier and no token URL will cause entries to be saved with the same key for bookmarks using same protocol but different OAuth endpoint.

dkocher: @AliveDevil The previous default was `duck:identifier:fqdn:port?user= `, not `duck:fqdn:port?user= `?

dkocher:
Refactored to protocol service `PasswordStorePrefixService`

- Default implementation implements current behaviour of `DefaultHostPasswordStore`
- New specific implementation for Windows implements current behaviour of `CredentialManagerPasswordStore`

core/src/main/java/ch/cyberduck/core/DefaultHostPasswordStore.java
```diff
@@ -214,25 +202,27 @@ public void save(final Host bookmark) throws LocalAccessDeniedException {
...
             if(StringUtils.isNotBlank(credentials.getOauth().getRefreshToken())) {
-                this.addPassword(bookmark.getProtocol().getScheme(),
-                        bookmark.getPort(), this.getOAuthHostname(bookmark),
+                this.addPassword(scheme, port, hostname,
                         String.format("%s OAuth2 Refresh Token", prefix), credentials.getOauth().getRefreshToken());
             }
             // Save expiry
             if(credentials.getOauth().getExpiryInMilliseconds() != null) {
-                this.addPassword(this.getOAuthHostname(bookmark), String.format("%s OAuth2 Token Expiry", prefix),
+                this.addPassword(hostname, String.format("%s OAuth2 Token Expiry", prefix),
```
chenkins, line 221: Why is expiry stored under `hostname` only, whereas refresh, access and ID tokens are stored with scheme and port? Comment on this in the source code.

core/src/main/java/ch/cyberduck/core/AbstractProtocol.java
```diff
@@ -407,6 +407,13 @@ public <T> T getFeature(final Class<T> type) {
         if(type == ComparisonService.class) {
             return (T) new DefaultComparisonService(this);
         }
+        if(type == PasswordStorePrefixService.class) {
```
chenkins, line 410: Implementation via protocol feature without configurable property for implementation forces sub-classing all used protocols, not full implementation of #15113. Re-introduce factory for `factory.passwordstoreprefixservice.class` (and second Windows one)?

core/src/main/java/ch/cyberduck/core/PasswordStorePrefixService.java
```diff
@@ -0,0 +1,54 @@
...
+     */
+    Integer getPort(Host bookmark);
+
+    /**
+     * Returns Scheme for PasswordStore to save entry with
+     */
+    Scheme getScheme(Host bookmark);
+
+    /**
+     * Retrieves username information for this OAuth bookmark
```
chenkins, line 51: Why `OAuth bookmark` -> just `bookmark`?

core/src/main/java/ch/cyberduck/core/preferences/Preferences.java
```diff
@@ -554,6 +554,7 @@ protected void setFactories() {
         this.setDefault("factory.connectiontimeout.class", DefaultConnectionTimeout.class.getName());
         this.setDefault("factory.authorizationcodeprovider.class", "ch.cyberduck.core.oauth.BrowserOAuth2AuthorizationCodeProvider");
         this.setDefault("factory.s3.pathcontainerservice.class", "ch.cyberduck.core.s3.S3PathContainerService");
+        this.setDefault("factory.passwordstoreprefixservice.class", DefaultPasswordStorePrefixService.class.getName());
```
chenkins, line 557: Obsolete? There is no such factory.

core/src/main/java/ch/cyberduck/core/DefaultPasswordStorePrefixService.java
```diff
@@ -0,0 +1,65 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ */
+
+import org.apache.commons.lang3.StringUtils;
+
+import java.net.URI;
+
+/**
+ * Default OAuth Prefix Service. Will just return fields from passed bookmark.
```
chenkins, line 23: `OAuth Prefix Service` -> `Password Store Prefix Service`

windows/src/main/csharp/ch/cyberduck/ui/core/preferences/ApplicationPreferences.cs
```diff
@@ -72,6 +73,7 @@ protected override void setFactories()
             this.setDefault("factory.locale.class", typeof(DictionaryLocale).AssemblyQualifiedName);
             this.setDefault("factory.dateformatter.class", typeof(UserDefaultsDateFormatter).AssemblyQualifiedName);
             this.setDefault("factory.passwordstore.class", typeof(PasswordStoreFacade).AssemblyQualifiedName);
+            this.setDefault("factory.passwordstoreprefixservice.class", typeof(CredentialManagerPasswordStorePrefixService).AssemblyQualifiedName);
```
chenkins, line 76: Obsolete? There is no such factory.

cli/src/main/csharp/ch/cyberduck/cli/WindowsTerminalPreferences.cs
```diff
@@ -81,6 +82,7 @@ protected override void setFactories()
             this.setDefault("factory.applicationfinder.class", typeof(ShellApplicationFinder).AssemblyQualifiedName);
             this.setDefault("factory.local.class", typeof(SystemLocal).AssemblyQualifiedName);
             this.setDefault("factory.passwordstore.class", typeof(PasswordStoreFacade).AssemblyQualifiedName);
+            this.setDefault("factory.passwordstoreprefixservice.class", typeof(CredentialManagerPasswordStorePrefixService).AssemblyQualifiedName);
```
chenkins, line 85: Obsolete? There is no such factory.

core/src/main/java/ch/cyberduck/core/PasswordStorePrefixService.java
```diff
@@ -0,0 +1,54 @@
...
+ * <pre>${application.container.name}${:getIdentifier()}${:getOptionalPort()${?user=getUsername()}</pre>
+ * For DefaultHostPasswordStore the format is
+ * <pre>${getScheme()}://${getDescription()} $EntryName@${getHostname()}:${getPort()}</pre>
+ * <p>GetDescription can return "Protocol Description", or "Protocol Description (${getUsername()})".</p>
+ * <p>EntryName can be "OAuth2 Access Token", "OAuth2 Refresh Token", "OIDC Id Token", "OAuth2 Token Expiry"</p>
+ * </p>
+ */
+public interface PasswordStorePrefixService {
+    /**
+     * Gets a description to insert as User-component into PasswordStore
```
chenkins, line 31: Why `User-component`?

core/src/main/java/ch/cyberduck/core/PasswordStorePrefixService.java
```diff
@@ -0,0 +1,54 @@
...
+     */
+    String getPrefix(Host bookmark);
+
+    /**
+     * Returns the OAuth configured hostname of this bookmark.
```
chenkins, line 36: Why `OAuth configured`?

core/src/main/java/ch/cyberduck/core/DefaultHostPasswordStore.java
```diff
@@ -262,21 +252,26 @@ public void delete(final Host bookmark) throws LocalAccessDeniedException {
...
+                this.deletePassword(scheme, port, hostname,
                         String.format("%s OAuth2 Access Token", prefix));
             }
             if(StringUtils.isNotBlank(credentials.getOauth().getRefreshToken())) {
-                this.deletePassword(protocol.getScheme(), bookmark.getPort(), this.getOAuthHostname(bookmark),
+                this.deletePassword(scheme, port, hostname,
                         String.format("%s OAuth2 Refresh Token", prefix));
             }
             // Save expiry
             if(credentials.getOauth().getExpiryInMilliseconds() != null) {
```
chenkins, line 270: Dito.

core/src/main/java/ch/cyberduck/core/AbstractProtocol.java
dkocher: We have no use case where the implementation must be replaced for more than one protocol implementation?

core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
```diff
@@ -86,12 +86,12 @@ public override void deletePassword(Scheme scheme, int port, string hostName, st
...
         {
-            var target = ToUri(bookmark);
-            var cred = WinCredentialManager.GetCredentials(target.AbsoluteUri);
+            var cred = FindCredentials(bookmark);
             if (!string.IsNullOrWhiteSpace(cred.Password))
             {
                 return cred.Password;
             }
+
             return base.findLoginPassword(bookmark);
```
dkocher, line 95: Because the base implementation will load the platform specific implementation of the prefix service this will cause a regression in failure finding deprecated store entries.

core/src/main/java/ch/cyberduck/core/CredentialManagerPasswordStoreDescriptorService.java
```diff
@@ -0,0 +1,65 @@
...
+import java.net.URI;
+
+/**
+ * Implementation for Windows Credentials Manager
+ */
+public class CredentialManagerPasswordStoreDescriptorService extends DefaultPasswordStoreDescriptorService {
+
+    @Override
+    public String getDescriptor(final Host bookmark) {
+        return bookmark.getProtocol().getIdentifier();
```
dkocher, line 29: Using the identifier will not allow sharing between different protocols.

dkocher: The requirement is only that this behaviour can be overridden by a different implementation.

core/src/main/java/ch/cyberduck/core/DefaultHostPasswordStore.java
```diff
@@ -36,22 +34,18 @@ public abstract class DefaultHostPasswordStore implements HostPasswordStore {
...
-            log.warn("No hostname given");
-            return null;
-        }
-        final Credentials credentials = bookmark.getCredentials();
-        if(StringUtils.isEmpty(credentials.getUsername())) {
-            log.warn("No username given");
-            return null;
-        }
+        final PasswordStoreDescriptorService service = bookmark.getProtocol().getFeature(PasswordStoreDescriptorService.class);
+        final String prefix = service.getDescriptor(bookmark);
```
ylangisc, line 38: Unused.

core/src/main/java/ch/cyberduck/core/DefaultPasswordStoreDescriptorService.java
ylangisc, line 1: Introduce variables, e.g. `URI.create(bookmark.getProtocol().getOAuthTokenUrl())` to avoid duplicate code which is not very readable as you always have to check if there is any difference to the previous usage.

core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
```diff
@@ -211,27 +221,37 @@ public override void save(Host bookmark)
 
         private static Uri ToUri(Host bookmark)
         {
-            var protocol = bookmark.getProtocol();
-            var credentials = bookmark.getCredentials();
-
-            var targetBuilder = new UriBuilder(PreferencesFactory.get().getProperty("application.container.name"), string.Empty);
+            PasswordStoreDescriptorService service = (PasswordStoreDescriptorService)bookmark.getProtocol().getFeature(typeof(PasswordStorePrefixService));
```
ylangisc, line 224: Wrong type `PasswordStorePrefixService`.

core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
```diff
@@ -211,27 +221,37 @@ public override void save(Host bookmark)
...
-                if (protocol.isPortConfigurable() && !Equals(protocol.getDefaultPort(), bookmark.getPort()))
-                {
-                    pathBuilder.Append(":" + bookmark.getPort());
-                }
+                pathBuilder.Append(prefix);
+            }
+            string hostname = service.getHostname(bookmark);
+            if (!string.IsNullOrWhiteSpace(hostname))
+            {
+                pathBuilder.Append(hostname);
```
ylangisc, line 235: Missing `:` between prefix and hostname?

core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
```diff
@@ -211,27 +221,37 @@ public override void save(Host bookmark)
 
         private static Uri ToUri(Host bookmark)
```
ylangisc, line 222: Missing test coverage for such methods.

core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
```diff
@@ -211,27 +221,37 @@ public override void save(Host bookmark)
...
         {
-            var protocol = bookmark.getProtocol();
-            var credentials = bookmark.getCredentials();
-
-            var targetBuilder = new UriBuilder(PreferencesFactory.get().getProperty("application.container.name"), string.Empty);
+            PasswordStoreDescriptorService service = (PasswordStoreDescriptorService)bookmark.getProtocol().getFeature(typeof(PasswordStorePrefixService));
+            var targetBuilder = new UriBuilder(new HostPreferences(bookmark).getProperty("application.container.name"), string.Empty);
             var pathBuilder = new StringBuilder();
-            pathBuilder.Append(protocol.getIdentifier());
-            if (protocol.isHostnameConfigurable() || !(protocol.isTokenConfigurable() || protocol.isOAuthConfigurable()))
```
ylangisc, line 220: Is this equivalent to the new checks in `CredentialManagerPasswordStoreDescriptorService` which are different.

PR #15138, opened 2023-09-23
Fix #15137

### LINKED ISSUE #15137: Provide preference to enable versioning of files edited in external application
With #14376 previous versions of files can be kept prior replacing on upload. An option should be provided in _Preferences → Editor_ to enable for files uploaded after editing in external application.

dkocher:
## macOS

image

## Windows

image


AliveDevil (changes requested):
Missing Preferences.strings.1 for WinForms.

ylangisc (approved):

AliveDevil (approved):

PR #15203, opened 2023-10-12
Update runtime dependency.

dkocher:


ylangisc (changes requested):
Update missing in `cli/osx/pom.xml`.

ylangisc (approved):

PR #15251, opened 2023-10-23
Fix #15250.

### LINKED ISSUE #15250: Application scoped bookmark to access file outside of sandbox not saved

dkocher:


ylangisc:
This has quite a high potential for regressions when refactoring things back like b8283badcd08b89815de5f9f611a2b3295980e51. Can we find a way to ensure that the default implementation isn't used?

ylangisc (approved):

PR #15256, opened 2023-10-24
Fix #15255.

### LINKED ISSUE #15255: Keeps putting copy of bookmarks in trash when launched.
**Describe the bug**
Mountain Duck keeps putting bookmarks in the trash upon launch if a connection was previously open prior to close.

**To Reproduce**
1. Enable "Save Workspace".
2. Setup and connect to any service. Save the bookmark.
3. With the connection "connected", quit Mountain Duck.
4. Relaunch Mountain Duck and within a few seconds, a copy of the stored bookmarks will be put in the trash.

**Expected behavior**
Mountain Duck should not be putting any bookmarks in the trash unless the user has chosen to delete the bookmark(s).

 - OS: MacOS Ventura
 - Version 13.3.1

**Log Files**
Search log file for "DELETE" and you will see that Mountain Duck seems to think a filesystem DELETE event is emitted.

**Additional context**
Can reproduce problem on both a physical Macbook Pro install as well as in a fresh MacOS VM. Using direct download version of Mountain Duck.
[mountainduck.log](https://github.com/iterate-ch/cyberduck/files/13109308/mountainduck.log)

dkocher:


core/src/main/java/ch/cyberduck/core/BookmarkCollection.java
```diff
@@ -91,6 +93,13 @@ public void collectionItemRemoved(final Host bookmark) {
                 log.debug("Skip indexing collection while loading");
             }
             else {
+                final Local file = this.getFile(bookmark);
+                try {
+                    LocalTrashFactory.get().trash(file);
```
ylangisc, line 98: Still trashes session bookmarks during start in https://github.com/iterate-ch/cyberduck/blob/3fef1fb760e93d545d1f5c4da110ffb26eab13a0/osx/src/main/java/ch/cyberduck/ui/cocoa/controller/MainController.java#L956

ylangisc (changes requested):

ylangisc (approved):

PR #15287, opened 2023-10-29
Update runtime dependency.

dkocher:


ylangisc (changes requested):
GitHub workflow files need to be adjusted too.

dkocher:
> GitHub workflow files need to be adjusted too.

The runtime is managed separately from the build JDK.

ylangisc:
> > GitHub workflow files need to be adjusted too.
> 
> The runtime is managed separately from the build JDK.

Referring to https://github.com/iterate-ch/cyberduck/blob/master/.github/workflows/build.yml#L21

ylangisc (approved):

PR #15388, opened 2023-11-22
Ensure returned aliases are unique

### LINKED ISSUE #15391: Don't show multiple same-labeled client certificates to select
The java KeyStore-API doesn't have a way to differentiate certificates[^1], thus all certificates with the same DN-value are considered identical.
[^1]: https://docs.oracle.com/javase/8/docs/api/java/security/KeyStore.html#getCertificate-java.lang.String-

To align with this behavior SystemCertificateStore should return a unique list of aliases.

AliveDevil:
Resolves #15391

### DIFF
--- a/core/src/main/csharp/ch/cyberduck/core/SystemCertificateStore.cs
+++ b/core/src/main/csharp/ch/cyberduck/core/SystemCertificateStore.cs
@@ -18,6 +18,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
 using System.Security.Cryptography.X509Certificates;
 using ch.cyberduck.core;
 using ch.cyberduck.core.exception;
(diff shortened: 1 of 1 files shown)


core/src/main/csharp/ch/cyberduck/core/SystemCertificateStore.cs
```diff
@@ -186,24 +187,21 @@ public bool verify(CertificateTrustCallback prompt, String hostName, List certs)
...
-                    certs.Add(String.IsNullOrEmpty(certificate.FriendlyName)
-                        ? certificate.GetNameInfo(X509NameType.SimpleName, false)
-                        : certificate.FriendlyName);
+                    Log.debug($"Skipping duplicate alias \"{alias}\"");
                 }
             }
-            finally
-            {
-                store.Close();
-            }
```
dkocher, line 206: Closing the store is no longer required?

AliveDevil: Using / Dispose-Pattern does this automatically.

AliveDevil:
Update description.

ylangisc (approved):

PR #15392, opened 2023-11-22
Add feature flags for write options.

dkocher:


core/src/test/java/ch/cyberduck/test/VaultTest.java
```diff
@@ -37,6 +38,7 @@ public static void credentials() {
         final VaultConfig config;
         try {
             config = new VaultConfig()
+                    .sslConfig(new SslConfig().verify(false))
```
ylangisc, line 41: Shouldn't be in the changeset?

ylangisc (changes requested):

ylangisc (approved):

PR #15393, opened 2023-11-22
Save OAuth tokens with port number, scheme and hostname from token URL.

### LINKED ISSUE #15113: Make OAuth prefix used for storing credentials configurable; use port from token URL instead from protocol

### LINKED ISSUE #15284: Shared OAuth refresh token not accepted

dkocher:
- Resolves #15113
- Resolves #15284

### DIFF
--- a/core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
+++ b/core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
@@ -60,8 +60,8 @@ namespace Ch.Cyberduck.Core
             {
                 logger.info(string.Format("Delete password for bookmark {0}", bookmark));
             }
-            var target = ToUri(bookmark);
-            foreach (Uri descriptor in target)
+            var targets = ToUri(bookmark);
+            foreach (Uri descriptor in targets)
             {
                 WinCredentialManager.RemoveCredentials(descriptor.AbsoluteUri);
             }
@@ -119,8 +119,8 @@ namespace Ch.Cyberduck.Core
             {
                 logger.info(string.Format("Fetching OAuth tokens from keychain for {0}", bookmark));
             }
-            var target = ToUri(bookmark);
-            foreach(Uri descriptor in target)
+            var targets = ToUri(bookmark);
+            foreach(Uri descriptor in targets)
             {
                 var cred = WinCredentialManager.GetCredentials(descriptor.AbsoluteUri);
                 if (cred.Attributes is Dictionary<string, string> attrs)
--- a/core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
+++ b/core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
@@ -140,9 +140,8 @@ namespace Ch.Cyberduck.Core
                     }
                     // Continue with deprecated descriptors
                 }
-                return base.findOAuthTokens(bookmark);
             }
-            return OAuthTokens.EMPTY;
+            return base.findOAuthTokens(bookmark);
         }
 
         public override string findPrivateKeyPassphrase(Host bookmark)
--- a/core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
+++ b/core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
@@ -21,11 +21,8 @@ using ch.cyberduck.core.exception;
 using ch.cyberduck.core.preferences;
 using Ch.Cyberduck.Core.CredentialManager;
 using org.apache.logging.log4j;
-using org.apache.logging.log4j.core.net;
 using System;
 using System.Collections.Generic;
-using System.Collections.ObjectModel;
-using System.Linq;
 using System.Net;
 using System.Text;
 using static Windows.Win32.Security.Credentials.CRED_PERSIST;
@@ -221,9 +218,9 @@ namespace Ch.Cyberduck.Core
             }
         }
 
-        private static Uri[] ToUri(Host bookmark)
+        private static List<Uri> ToUri(Host bookmark)
         {
-            Collection<Uri> descriptors = new();
+            List<Uri> descriptors = new();
             foreach(string descriptor in ToDescriptor(bookmark))
             {
                 var protocol = bookmark.getProtocol();
@@ -247,7 +244,7 @@ namespace Ch.Cyberduck.Core
                 }
                 descriptors.Add(targetBuilder.Uri);
             }
-            return descriptors.ToArray();
+            return descriptors;
         }
 
         private static string[] ToDescriptor(Host bookmark)
--- a/core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
+++ b/core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
@@ -66,11 +66,9 @@ namespace Ch.Cyberduck.Core
             var target = ToUri(bookmark);
             foreach (Uri descriptor in target)
             {
-                if (!WinCredentialManager.RemoveCredentials(descriptor.AbsoluteUri))
-                {
-                    base.delete(bookmark);
-                }
+                WinCredentialManager.RemoveCredentials(descriptor.AbsoluteUri);
             }
+            base.delete(bookmark);
         }
 
         public override void deletePassword(string serviceName, string user)
@@ -145,7 +143,6 @@ namespace Ch.Cyberduck.Core
                     }
                     // Continue
                 }
-
                 return base.findOAuthTokens(bookmark);
             }
             return OAuthTokens.EMPTY;
--- a/core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
+++ b/core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
@@ -21,8 +21,11 @@ using ch.cyberduck.core.exception;
 using ch.cyberduck.core.preferences;
 using Ch.Cyberduck.Core.CredentialManager;
 using org.apache.logging.log4j;
+using org.apache.logging.log4j.core.net;
 using System;
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Linq;
 using System.Net;
 using System.Text;
 using static Windows.Win32.Security.Credentials.CRED_PERSIST;
@@ -61,9 +64,12 @@ namespace Ch.Cyberduck.Core
                 logger.info(string.Format("Delete password for bookmark {0}", bookmark));
             }
             var target = ToUri(bookmark);
-            if (!WinCredentialManager.RemoveCredentials(target.AbsoluteUri))
+            foreach (Uri descriptor in target)
             {
-                base.delete(bookmark);
+                if (!WinCredentialManager.RemoveCredentials(descriptor.AbsoluteUri))
+                {
+                    base.delete(bookmark);
+                }
             }
         }
 
@@ -86,7 +92,7 @@ namespace Ch.Cyberduck.Core
 
         public override string findLoginPassword(Host bookmark)
         {
-            var target = ToUri(bookmark);
+            var target = ToUri(bookmark)[0];
             var cred = WinCredentialManager.GetCredentials(target.AbsoluteUri);
             if (!string.IsNullOrWhiteSpace(cred.Password))
             {
@@ -101,7 +107,7 @@ namespace Ch.Cyberduck.Core
             {
                 logger.info(string.Format("Fetching login token from keychain for {0}", bookmark));
             }
-            var target = ToUri(bookmark);
+            var target = ToUri(bookmark)[0];
             var cred = WinCredentialManager.GetCredentials(target.AbsoluteUri);
             if (cred.Attributes is Dictionary<string, string> attrs
                 && attrs.TryGetValue("Token", out var token)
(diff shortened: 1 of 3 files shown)


core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
```diff
@@ -216,7 +216,14 @@ private static Uri ToUri(Host bookmark)
 
             var targetBuilder = new UriBuilder(PreferencesFactory.get().getProperty("application.container.name"), string.Empty);
             var pathBuilder = new StringBuilder();
-            pathBuilder.Append(protocol.getIdentifier());
+            if (protocol.isOAuthConfigurable())
+            {
+                pathBuilder.Append(protocol.getOAuthClientId());
```
dkocher, line 221: Will need to add fallback to use previous identifier in lookup.

core/src/main/java/ch/cyberduck/core/DefaultHostPasswordStore.java
```diff
@@ -174,11 +181,27 @@ protected String getOAuthHostname(final Host bookmark) {
         return bookmark.getHostname();
     }
 
-    private String getOAuthPrefix(final Host bookmark) {
-        if(StringUtils.isNotBlank(bookmark.getCredentials().getUsername())) {
-            return String.format("%s (%s)", bookmark.getProtocol().getDescription(), bookmark.getCredentials().getUsername());
+    protected int getOAuthPort(final Host bookmark) {
+        if(-1 != URI.create(bookmark.getProtocol().getOAuthTokenUrl()).getPort()) {
```
AliveDevil, line 185: Reuse URI.create-instance.

core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
```diff
@@ -209,29 +224,42 @@ public override void save(Host bookmark)
             }
         }
 
-        private static Uri ToUri(Host bookmark)
+        private static Uri[] ToUri(Host bookmark)
```
AliveDevil, line 227: Use `List<Uri>`.
It is in private use, doesn't need allocating another array from a Collection.

core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
```diff
@@ -61,9 +64,12 @@ public override void delete(Host bookmark)
                 logger.info(string.Format("Delete password for bookmark {0}", bookmark));
             }
             var target = ToUri(bookmark);
-            if (!WinCredentialManager.RemoveCredentials(target.AbsoluteUri))
+            foreach (Uri descriptor in target)
             {
-                base.delete(bookmark);
+                if (!WinCredentialManager.RemoveCredentials(descriptor.AbsoluteUri))
+                {
+                    base.delete(bookmark);
```
AliveDevil, line 71: This will get called multiple times.
Unconditionally call base.delete after loop then.

core/src/main/java/ch/cyberduck/core/DefaultHostPasswordStore.java
```diff
@@ -148,23 +148,30 @@ public OAuthTokens findOAuthTokens(final Host bookmark) {
...
-        catch(LocalAccessDeniedException e) {
-            log.warn(String.format("Failure %s searching in keychain", e));
-            return OAuthTokens.EMPTY;
+        final String[] descriptors = this.getOAuthPrefix(bookmark);
+        for(String prefix : descriptors) {
+            final String hostname = this.getOAuthHostname(bookmark);
+            try {
+                final String expiry = this.getPassword(this.getOAuthHostname(bookmark), String.format("%s OAuth2 Token Expiry", prefix));
+                final OAuthTokens tokens = new OAuthTokens(
+                        this.getPassword(bookmark.getProtocol().getScheme(), this.getOAuthPort(bookmark), hostname,
```
chenkins, line 157: Add `getOAuthScheme()`, otherwise port comes from token URL, but scheme comes from bookmark's protocol.

chenkins (changes requested):

AliveDevil (approved):
Looks good to me.

core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
```diff
@@ -119,21 +120,29 @@ public override OAuthTokens findOAuthTokens(Host bookmark)
...
+                    if (attrs.TryGetValue("OAuth Expiry", out var expiryValue))
+                    {
+                        long.TryParse(expiryValue, out expiry);
+                    }
+                    OAuthTokens tokens = new(accessToken, refreshToken, new(expiry), idToken);
+                    if(tokens.validate())
+                    {
+                        return tokens;
+                    }
+                    // Continue
```
chenkins, line 141:
```suggestion
                    // continue: fallback to old behaviour for backwards compatibility
```

core/src/main/java/ch/cyberduck/core/DefaultHostPasswordStore.java
```diff
@@ -204,36 +235,39 @@ public void save(final Host bookmark) throws LocalAccessDeniedException {
...
+                if(credentials.getOauth().getExpiryInMilliseconds() != null) {
+                    this.addPassword(getOAuthHostname(bookmark), String.format("%s OAuth2 Token Expiry", prefix),
+                            String.valueOf(credentials.getOauth().getExpiryInMilliseconds()));
+                }
+                if(StringUtils.isNotBlank(credentials.getOauth().getIdToken())) {
+                    this.addPassword(getOAuthScheme(bookmark),
+                            getOAuthPort(bookmark), getOAuthHostname(bookmark),
+                            String.format("%s OIDC Id Token", prefix), credentials.getOauth().getIdToken());
+                }
+                break;
```
chenkins, line 270:
```suggestion
                break; // new behaviour
```

core/src/main/java/ch/cyberduck/core/DefaultHostPasswordStore.java
```diff
@@ -204,36 +235,39 @@ public void save(final Host bookmark) throws LocalAccessDeniedException {
...
-            if(credentials.getOauth().getExpiryInMilliseconds() != null) {
-                this.addPassword(this.getOAuthHostname(bookmark), String.format("%s OAuth2 Token Expiry", prefix),
-                        String.valueOf(credentials.getOauth().getExpiryInMilliseconds()));
-            }
-            if(StringUtils.isNotBlank(credentials.getOauth().getIdToken())) {
-                this.addPassword(bookmark.getProtocol().getScheme(),
-                        bookmark.getPort(), this.getOAuthHostname(bookmark),
-                        String.format("%s OIDC Id Token", prefix), credentials.getOauth().getIdToken());
+            final String[] descriptors = getOAuthPrefix(bookmark);
+            for(String prefix : descriptors) {
```
chenkins, line 249:
```suggestion
            final String prefix = getOAuthPrefix(bookmark)[0]; // new behaviour
```

core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
```diff
@@ -172,7 +181,7 @@ public override void save(Host bookmark)
             {
                 logger.info(string.Format("Add password for bookmark {0}", bookmark));
             }
-            var target = ToUri(bookmark);
+            var target = ToUri(bookmark)[0];
```
chenkins, line 184:
```suggestion
            var target = ToUri(bookmark)[0]; // new behaviour
```

chenkins (commented):
Left my comments/suggestions.

core/src/main/java/ch/cyberduck/core/DefaultHostPasswordStore.java
```diff
@@ -148,37 +148,67 @@ public OAuthTokens findOAuthTokens(final Host bookmark) {
...
+            return uri.getPort();
         }
-        return bookmark.getProtocol().getDescription();
+        return getOAuthScheme(bookmark).getPort();
+    }
+
+    private static String[] getOAuthPrefix(final Host bookmark) {
+        if(StringUtils.isNotBlank(bookmark.getCredentials().getUsername())) {
+            return new String[]{
+                    String.format("%s (%s)", bookmark.getProtocol().getOAuthClientId(), bookmark.getCredentials().getUsername()),
```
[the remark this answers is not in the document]
dkocher: See 1f257ff200d4d98585772cddacd258898d6b52bd.

core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
```diff
@@ -142,7 +151,7 @@ public override string findPrivateKeyPassphrase(Host bookmark)
             {
                 logger.info(string.Format("Fetching private key passphrase from keychain for {0}", bookmark));
             }
-            var target = ToUri(bookmark);
+            var target = ToUri(bookmark)[0];
```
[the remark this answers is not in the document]
dkocher: The implementation of `ToUri(Host bookmark)` is supposed to return a single element only for non OAuth bookmarks equal to current.

core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
```diff
@@ -86,7 +87,7 @@ public override void deletePassword(Scheme scheme, int port, string hostName, st
 
         public override string findLoginPassword(Host bookmark)
         {
-            var target = ToUri(bookmark);
+            var target = ToUri(bookmark)[0];
```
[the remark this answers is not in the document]
dkocher: The implementation of `ToUri(Host bookmark)` is supposed to return a single element only for non OAuth bookmarks equal to current.

core/src/main/java/ch/cyberduck/core/DefaultHostPasswordStore.java
```diff
@@ -148,37 +148,67 @@ public OAuthTokens findOAuthTokens(final Host bookmark) {
...
+    private static int getOAuthPort(final Host bookmark) {
+        final URI uri = URI.create(bookmark.getProtocol().getOAuthTokenUrl());
+        if(-1 != uri.getPort()) {
+            return uri.getPort();
         }
-        return bookmark.getProtocol().getDescription();
+        return getOAuthScheme(bookmark).getPort();
+    }
+
+    private static String[] getOAuthPrefix(final Host bookmark) {
```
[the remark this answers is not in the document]
dkocher: Agree we could have this explicit instead of convention. But this version is easily expandable for future changes by adding a new prefix in [0]

core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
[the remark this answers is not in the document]
AliveDevil: FindLoginPassword could should check for all possible versions _for reading_. Saving should be handled for the current Uri-format only.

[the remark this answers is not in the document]
AliveDevil: Looking at the implementation - well, there should only be one password assigned.

core/src/main/java/ch/cyberduck/core/DefaultHostPasswordStore.java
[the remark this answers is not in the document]
chenkins: ©me 🤦 
@dkocher Does this pr fix #15284? If yes, add `Fixes  #15284`.

core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
AliveDevil: One nitpick here:
Yes, fall back, but shouldn't _all_ OAuth Uris be tried?
This checks the first, and then returns to base-implementation - which should regardless.

dkocher: Good catch. Should only fallback to base implementation after trying all descriptors.

AliveDevil: Slipped through the first time looking at it.

core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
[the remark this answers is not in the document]
chenkins: Missed the case distinction in `ToDescriptor`.  

So as `findLoginPassword` and `findLoginToken` and `findPrivateKeyPassphrase` are only used in the non-OAuth case and `ToUri` returns the old behaviour for them, it's correct....

core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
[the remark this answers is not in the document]
chenkins: > Looking at the implementation - well, there should only be one password assigned.

Should - famous last words 😉

core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
```diff
@@ -142,7 +150,7 @@ public override string findPrivateKeyPassphrase(Host bookmark)
             {
                 logger.info(string.Format("Fetching private key passphrase from keychain for {0}", bookmark));
             }
-            var target = ToUri(bookmark);
+            var target = ToUri(bookmark)[0];
```
chenkins, line 153:
```suggestion
            var target = ToUri(bookmark)[0]; // non-OAuth case
```

core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
```diff
@@ -172,7 +180,7 @@ public override void save(Host bookmark)
             {
                 logger.info(string.Format("Add password for bookmark {0}", bookmark));
             }
-            var target = ToUri(bookmark);
+            var target = ToUri(bookmark)[0];
```
chenkins, line 183:
```suggestion
            var target = ToUri(bookmark)[0]; ToUri(bookmark)[0]; // non-OAuth case
```

core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
chenkins, line 90:
```suggestion
            var target = ToUri(bookmark)[0]; // non-OAuth case
```

core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
```diff
@@ -209,29 +217,42 @@ public override void save(Host bookmark)
...
-            if (!string.IsNullOrWhiteSpace(credentials.getUsername()))
+            return descriptors;
+        }
+
+        private static string[] ToDescriptor(Host bookmark)
+        {
+            if(bookmark.getProtocol().isOAuthConfigurable())
             {
-                targetBuilder.Query = "user=" + credentials.getUsername();
+                return new string[] { bookmark.getProtocol().getOAuthClientId(), bookmark.getProtocol().getIdentifier() };
```
chenkins, line 253:
```suggestion
                return new string[] { bookmark.getProtocol().getOAuthClientId(), bookmark.getProtocol().getIdentifier() }; // OAuth case with backwards compatibility
```

core/src/main/csharp/ch/cyberduck/core/CredentialManagerPasswordStore.cs
```diff
@@ -209,29 +217,42 @@ public override void save(Host bookmark)
...
             }
-
-            return targetBuilder.Uri;
+            return new string[] { bookmark.getProtocol().getIdentifier() };
```
chenkins, line 255:
```suggestion
            return new string[] { bookmark.getProtocol().getIdentifier() }; // non-OAuth case (no backwards compatibility required)
```

chenkins (approved):
LGTM with suggestions.

AliveDevil (approved):

ylangisc (approved):

PR #15398, opened 2023-11-24
Update dependencies

dkocher:


pom.xml
```diff
@@ -131,6 +131,11 @@
                     </exclusion>
                 </exclusions>
             </dependency>
+            <dependency>
+                <groupId>com.jamesmurty.utils</groupId>
```
ylangisc, line 135: Should be updated here: https://github.com/iterate-ch/jets3t/blob/master/pom.xml#L319

pom.xml
```diff
@@ -403,6 +408,16 @@
                 <artifactId>jaxb-runtime</artifactId>
                 <version>2.3.8</version>
             </dependency>
+            <dependency>
```
ylangisc, line 411: Kotlin should be updated here: https://github.com/iterate-ch/ds3_java_sdk/blob/develop/pom.xml#L30

ylangisc (changes requested):

pom.xml
ylangisc: Managed libs can be removed as sdk version has been updated?

ylangisc (changes requested):

ylangisc (approved):

PR #15461, opened 2023-12-13
Handle 5xx as offline.

dkocher:


core/src/main/java/ch/cyberduck/core/diagnostics/HttpReachability.java
```diff
@@ -79,6 +80,15 @@ public boolean isReachable(final Host bookmark) {
                 log.debug(String.format("Received response %s", response));
             }
             EntityUtils.consume(response.getEntity());
+            switch(response.getStatusLine().getStatusCode()) {
+                case HttpStatus.SC_BAD_GATEWAY:
+                case HttpStatus.SC_INTERNAL_SERVER_ERROR:
+                case HttpStatus.SC_SERVICE_UNAVAILABLE:
```
ylangisc, line 86: Add `SC_GATEWAY_TIMEOUT`?

ylangisc (approved):

PR #15512, opened 2024-01-05
Throw failure when failing to open temporary file for writing.

dkocher:


core/src/main/java/ch/cyberduck/core/io/FileBuffer.java
```diff
@@ -35,21 +35,27 @@ public class FileBuffer implements Buffer {
...
-    public FileBuffer() {
+    public FileBuffer() throws AccessDeniedException {
         this(TemporaryFileServiceFactory.get().create(new AlphanumericRandomStringService().random()));
     }
 
-    public FileBuffer(final Local temporary) {
+    public FileBuffer(final Local temporary) throws AccessDeniedException {
         this.temporary = temporary;
+        try {
+            this.file = new RandomAccessFile(temporary.getAbsolute(), "rw");
```
ylangisc, line 49: Most probably not a good idea to create a `RandomAccessFile` in the constructor as in Mountain Duck for every open operation a new `FileBuffer` is created.

ylangisc (changes requested):

PR #15516, opened 2024-01-05
Add mapping for permissions.

dkocher:


brick/src/main/java/ch/cyberduck/core/brick/BrickAttributesFinderFeature.java
```diff
@@ -75,6 +76,29 @@ public PathAttributes toAttributes(final FileEntity entity) {
         else if(entity.getMtime() != null) {
             attr.setModificationDate(entity.getMtime().getMillis());
         }
+        if(entity.getPermissions() != null) {
+            final Permission permission = new Permission();
+            switch(PermissionType.valueOf(entity.getPermissions())) {
+                case readonly:
+                case list:
```
ylangisc, line 83: Missing `execute` for `list` and folders?

ylangisc (approved):

PR #15517, opened 2024-01-05
Migrate files from Packaged Location to shared location

### LINKED ISSUE #15514: Migrate files from Packaged Location to shared location
Files in packaged location (previously packaged redirection for `FOLDERID_RoamingAppData` and `FOLDERID_LocalAppData`) are now inaccessible by Cyberduck (by normal means).
Perform migration from packaged `FOLDERID_RoamingAppData` to unpackaged location.

Fixes #15457

AliveDevil:
Fixes #15514

### DIFF
--- a/windows/src/main/csharp/ch/cyberduck/ui/controller/MainController.cs
+++ b/windows/src/main/csharp/ch/cyberduck/ui/controller/MainController.cs
@@ -60,6 +60,7 @@ using ch.cyberduck.core.urlhandler;
 using Ch.Cyberduck.Core;
 using Ch.Cyberduck.Core.Sparkle;
 using Ch.Cyberduck.Core.TaskDialog;
+using Ch.Cyberduck.Core.Ui.Preferences;
 using Ch.Cyberduck.Ui.Core.Contracts;
 using java.util;
 using org.apache.logging.log4j;
@@ -879,12 +880,25 @@ namespace Ch.Cyberduck.Ui.Controller
             }
         }
 
-        [MethodImpl(MethodImplOptions.NoInlining)]
         private void InitStoreContext()
         {
             var storeContext = StoreContext.GetDefault();
             var initWindow = (IInitializeWithWindow)(object)storeContext;
             initWindow.Initialize((HWND)MainForm.Handle);
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private void OnPackagedStartup()
+        {
+            InitStoreContext();
+            try
+            {
+                PackagedDataMigrator.Migrate();
+            }
+            catch (Exception e)
+            {
+                Logger.warn("Uncaught failure in data migration", e);
+            }
         }
 
         private void OnStartup(object state)
@@ -895,7 +909,7 @@ namespace Ch.Cyberduck.Ui.Controller
             /* UWP Registration, initialize as soon as possible */
             if (Utils.IsRunningAsUWP)
             {
-                InitStoreContext();
+                OnPackagedStartup();
             }
 
             InitializeTransfers();
--- a/windows/src/main/csharp/ch/cyberduck/ui/core/preferences/PackagedDataMigrator.cs
+++ b/windows/src/main/csharp/ch/cyberduck/ui/core/preferences/PackagedDataMigrator.cs
@@ -1,0 +1,97 @@
+﻿using ch.cyberduck.core.preferences;
+using Ch.Cyberduck.Core.Preferences;
+using org.apache.logging.log4j;
+using System;
+using System.IO;
+using Windows.Win32.UI.Shell;
+using static Windows.Win32.CorePInvoke;
+using static Windows.Win32.PInvoke;
+
+namespace Ch.Cyberduck.Core.Ui.Preferences;
+
+internal static class PackagedDataMigrator
+{
+    public static void Migrate()
+    {
+        // Logger is used for short time.
+        // Don't need to keep an instance alive longer than necessary
+        var logger = LogManager.getLogger(typeof(PackagedDataMigrator).FullName);
+        var preferences = PreferencesFactory.get();
+        if (preferences.getBoolean("application.datadir.migrate"))
+        {
+            if (logger.isDebugEnabled())
+            {
+                logger.debug("Skipping DataDir migration");
+            }
+
+            return;
+        }
+
+        if (FindPackagedPath(logger) is { } path && !string.IsNullOrWhiteSpace(path))
+        {
+            var targetPath = RoamingSupportDirectoryFinder.Local.getAbsolute();
+            foreach (var item in new[] { "Bookmarks", "History", "Profiles", "Sessions", "Transfers" })
+            {
+                DirectoryInfo directory = new(Path.Combine(path, item));
+                if (!directory.Exists)
+                {
+                    continue;
+                }
+
+                var copyTarget = Path.Combine(targetPath, item);
+                foreach (var file in directory.EnumerateFiles())
+                {
+                    FileInfo targetFile = new(Path.Combine(copyTarget, file.Name));
+                    if (targetFile.Exists && targetFile.LastWriteTime >= file.LastWriteTime)
+                    {
+                        if (logger.isInfoEnabled())
+                        {
+                            logger.info($"Skipping {item}/{file.Name} - Exists and target is newer");
+                        }
+
+                        continue;
+                    }
+
+                    try
+                    {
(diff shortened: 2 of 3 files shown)


dkocher (changes requested):
Squash commits.

windows/src/main/csharp/ch/cyberduck/ui/core/preferences/PackagedDataMigrator.cs
```diff
@@ -0,0 +1,102 @@
...
+                }
+                finally
+                {
+                    if (!failedMigrate)
+                    {
+                        // skip moving when file couldn't be copied to shared lcoation
+                        try
+                        {
+                            // Move with overwrite
+                            file.CopyTo(Path.Combine(migratedDirectory.FullName, file.Name), true);
```
dkocher, line 93: Why not trash after migrating?

windows/src/main/csharp/ch/cyberduck/ui/core/preferences/PackagedDataMigrator.cs
```diff
@@ -0,0 +1,102 @@
...
+                    {
+                        // already migrated, or newer on shared location
+                        // move file to migrated-location
+
+                        if (logger.isInfoEnabled())
+                        {
+                            logger.info($"Skipping {item}/{file.Name} - Exists and target is newer");
+                        }
+
+                        continue;
```
dkocher, line 72: Finally block is executed regardless.

windows/src/main/csharp/ch/cyberduck/ui/core/preferences/PackagedDataMigrator.cs
```diff
@@ -0,0 +1,102 @@
...
+        var targetPath = RoamingSupportDirectoryFinder.Local.getAbsolute();
+        foreach (var item in new[] { "Bookmarks", "History", "Profiles", "Sessions", "Transfers" })
+        {
+            DirectoryInfo directory = new(Path.Combine(path, item));
+            if (!directory.Exists)
+            {
+                continue;
+            }
+
+            DirectoryInfo migratedDirectory = new(Path.Combine(migratedLocation.FullName, item));
```
dkocher, line 53: Scope too broad.

dkocher (changes requested):

windows/src/main/csharp/ch/cyberduck/ui/core/preferences/PackagedDataMigrator.cs
AliveDevil: Right, because following cases are covered:
- Copy from Packaged to Shared Location, but file exists in Shared Location already
  - Don't try to copy that file _every single time_ this migration is performed
- Copy from Packaged to Shared Location, but copy process failed
  - Don't move the migration-file, and try again next time
- Copy from Packaged to Shared Location, but file exists in Shared Location, and is older
  - Only move file away in case of successfully copying to shared location

windows/src/main/csharp/ch/cyberduck/ui/core/preferences/PackagedDataMigrator.cs
AliveDevil: Didn't want to delete files, that failed copying over to shared location - thus risking data loss.
Trashing might be an alternative.

windows/src/main/csharp/ch/cyberduck/ui/core/preferences/PackagedDataMigrator.cs
AliveDevil: Explain.

windows/src/main/csharp/ch/cyberduck/ui/core/preferences/PackagedDataMigrator.cs
AliveDevil: Exists in Shared, and timestamp equal => move to Migrated location
Exists in Shared, packaged newer => move to Migrated location if successful
Doesn't exist in Shared => move to Migrated location if successful

If copying to Shared location fails, don't move file to Migrated location, thus retrying the next time Cyberduck is started.

windows/src/main/csharp/ch/cyberduck/ui/controller/MainController.cs
```diff
@@ -879,14 +880,27 @@ private void InitializeUpdater()
...
+        private void OnPackagedStartup()
+        {
+            InitStoreContext();
+            try
+            {
+                PackagedDataMigrator.Migrate();
+            }
+            catch (Exception e)
+            {
+                Logger.warn("Uncatched failure in data migration", e);
```
ylangisc, line 900: ...Uncaught

ylangisc (commented):
Would implement this as a one-time, best-effort operation and set a property afterwards to be able to skip it. Just like we do with the 3rd party bookmark importers.

windows/src/main/csharp/ch/cyberduck/ui/core/preferences/PackagedDataMigrator.cs
```diff
@@ -0,0 +1,94 @@
...
+        {
+            if (logger.isDebugEnabled())
+            {
+                logger.debug("Skipping DataDir migration");
+            }
+
+            return;
+        }
+
+        preferences.setProperty("application.datadir.migrate", true);
```
dkocher, line 30: Set only *after* migrating.

dkocher (changes requested):

windows/src/main/csharp/ch/cyberduck/ui/core/preferences/PackagedDataMigrator.cs
AliveDevil: "best-effort" for me is try once, if it fails for whatever reason, don't try again later.

dkocher (approved):

AliveDevil:
With 72b4914 semantics changes, when:
- container path "%AppData%" doesn't exist
- or "%AppData%\Cyberduck` doesn't exist
migration property won't be set - thus always retried until end of time.

AliveDevil:
Reviewed,
when failing to get existent path (Packages\id\LocalCache\Roaming can't be retrieved, or PackagedAppData\Cyberduck doesn't exist) assume success, otherwise when failing unexpectedly retry migration next time

windows/src/main/csharp/ch/cyberduck/ui/core/preferences/PackagedDataMigrator.cs
```diff
@@ -0,0 +1,100 @@
...
+
+namespace Ch.Cyberduck.Core.Ui.Preferences;
+
+internal static class PackagedDataMigrator
+{
+    public static void Migrate()
+    {
+        // Logger is used for short time.
+        // Don't need to keep an instance alive longer than necessary
+        var logger = LogManager.getLogger(typeof(PackagedDataMigrator).FullName);
```
dkocher, line 18: Garbage collector will do its job instead of passing as parameter?

windows/src/main/csharp/ch/cyberduck/ui/core/preferences/PackagedDataMigrator.cs
```diff
@@ -0,0 +1,100 @@
...
+                }
+
+                var copyTarget = Path.Combine(targetPath, item);
+                foreach (var file in directory.EnumerateFiles())
+                {
+                    FileInfo targetFile = new(Path.Combine(copyTarget, file.Name));
+                    if (targetFile.Exists && targetFile.LastWriteTime >= file.LastWriteTime)
+                    {
+                        // already migrated, or newer on shared location
+                        // move file to migrated-location
```
dkocher, line 48: Delete out-of-date comment.

dkocher (changes requested):

windows/src/main/csharp/ch/cyberduck/ui/core/preferences/PackagedDataMigrator.cs
AliveDevil: I don't want to have the logger hanging around straining Gen2 (long living objects).
Same applies to java: static fields are _never_ collected by GC.

AliveDevil:
```
static class A
{
    static readonly object B = new();
}
```

B won't be collected until application shutdown.

dkocher (approved):

ylangisc (approved):

PR #15545, opened 2024-01-16
Vault factory for path only.

chenkins:


dkocher (approved):

ylangisc (commented):
The `null` parameters when calling the factory could be hidden by adding a second getter to the factory.

dkocher (changes requested):
As @ylangisc suggested it would be better to provide an additional factory method with different parameters. Would return `Vault.DISABLED` or throw `FactoryException` when the implementation does not support the constructor.

dkocher (approved):

ylangisc (approved):

PR #15600, opened 2024-02-01
Review descriptions.

dkocher:


ylangisc (commented):
Font change from `menu` to `system` is correct?

PR #15635, opened 2024-02-14
Configure rate limiting by default.

dkocher:


defaults/src/main/resources/default.properties
```diff
@@ -529,6 +531,7 @@ smb.domain.default=WORKGROUP
 # Enable distributed filesystem path resolver
 smb.dfs.enable=true
 # Requires that messages from the server are signed
+smb.encrypt.enable=false
```
AliveDevil, line 534: Not related?

ylangisc (commented):
Limiter should be aligned to the same metric as the backend which is based on requests per minute, not per second. E.g. a burst of 80 requests would require 10 seconds with a per second limiter whilst there is no need to limit the burst in the context of 500 requests per minute.

On the other hand if we use up the 500 requests within 10 seconds for example, we are blocked for almost 50 seconds until we can fire the next requests. What is a good trade-off to handle burst efficiently without long blocking times.

ylangisc (approved):

PR #15788, opened 2024-04-02
Protocol for resumable uploads implementation

### LINKED ISSUE #15619: Support resumable file uploads
> Infinite Scale supports the tus resumable-upload protocol [^1] 

[^1]: https://owncloud.dev/apis/http/tus_upload/

> tus is a project aiming to make resumable file uploads easily usable and widely available. [^2]

[^2]: https://tus.io/faq

dkocher:


owncloud/src/main/java/ch/cyberduck/core/owncloud/OwncloudSession.java
```diff
@@ -57,23 +62,63 @@
...
+        else {
+            upload = new HttpUploadFeature(new NextcloudWriteFeature(this));
+        }
+        return client;
+    }
+
+    private TusCapabilities options(final DAVClient client) throws BackgroundException {
+        final HttpOptions options = new HttpOptions(URIEncoder.encode(
+                new DelegatingHomeFeature(new DefaultPathHomeFeature(host)).find().getAbsolute()));
+        options.setHeader(TUS_HEADER_RESUMABLE, TUS_VERSION);
```
ylangisc, line 110: Not necessary for the `OPTIONS` request.
> The Tus-Resumable header MUST be included in every request and response except for OPTIONS requests

tus/src/main/java/ch/cyberduck/core/tus/TusCapabilitiesResponseHandler.java
```diff
@@ -0,0 +1,66 @@
...
+    @Override
+    public TusCapabilities handleResponse(final HttpResponse response) {
+        switch(response.getStatusLine().getStatusCode()) {
+            case HttpStatus.SC_NO_CONTENT:
+                if(log.isDebugEnabled()) {
+                    log.debug(String.format("Support for resumable uploads detected from %s", response));
+                }
+                final TusCapabilities tus = new TusCapabilities();
+                if(response.containsHeader(TusCapabilities.TUS_HEADER_RESUMABLE)) {
+                    tus.withVersion(response.getFirstHeader(TusCapabilities.TUS_HEADER_RESUMABLE).getValue());
```
ylangisc, line 39: Would rather try to match a version in `Tus-Version` with the one we support.

tus/src/main/java/ch/cyberduck/core/tus/TusUploadFeature.java
```diff
@@ -0,0 +1,243 @@
...
+                    }
+                    // If the resource is not found, the Server SHOULD return either the 404 Not Found,
+                    // 410 Gone or 403 Forbidden status without the Upload-Offset header.
+                    throw new HttpResponseException(response.getStatusLine().getStatusCode(), response.getStatusLine().getReasonPhrase());
+                }
+            });
+            return new Write.Append(true).withStatus(status).withSize(offset);
+        }
+        catch(HttpResponseException e) {
+            preferences.deleteProperty(property);
```
ylangisc, line 225: Check for status code 404, 410 or 403.

tus/src/main/java/ch/cyberduck/core/tus/TusCapabilitiesResponseHandler.java
dkocher: This is up to the caller [^1] 

[^1]: https://github.com/iterate-ch/cyberduck/pull/15788/files#diff-0ef80a78b96e7846613a681f29f6b14f6c2faf83a480a76e91a908bf8cfec80fR98

tus/src/main/java/ch/cyberduck/core/tus/TusUploadFeature.java
dkocher: Would rather not depend on. The tus implementation of ownCloud returns arbitrary status codes such as `502`.

owncloud/src/main/java/ch/cyberduck/core/owncloud/OwncloudSession.java
dkocher: I read that it in the sense that it *CAN* be included and thus we see the reason for the `OPTIONS` request.

tus/src/main/java/ch/cyberduck/core/tus/TusCapabilitiesResponseHandler.java
ylangisc: Of course but not the point. The following `OPTIONS` response would not enable the TUS extension although the server supports version 1.0.0:
```
HTTP/1.1 204 No Content
Tus-Resumable: 1.1.0
Tus-Version: 1.1.0,1.0.0,0.2.2,0.2.1
Tus-Max-Size: 1073741824
Tus-Extension: creation,expiration
```

owncloud/src/main/java/ch/cyberduck/core/owncloud/OwncloudSession.java
ylangisc: Don't get the point to add anything that is not needed and probably not evaluated on the servers side at this point.

tus/src/main/java/ch/cyberduck/core/tus/TusUploadFeature.java
```diff
@@ -0,0 +1,243 @@
...
+            final String uploadUrl;
+            if(status.isAppend()) {
+                uploadUrl = preferences.getProperty(toProperty(host, file));
+                if(log.isDebugEnabled()) {
+                    log.debug(String.format("Resume upload to %s for %s from offset %d", uploadUrl, file, status.getOffset()));
+                }
+            }
+            else {
+                // Create an Upload URL
+                final HttpPost request = new HttpPost(new DefaultUrlProvider(host).toUrl(file.getParent()).find(DescriptiveUrl.Type.provider).getUrl());
```
ylangisc, line 105: Usage of creation extension which is only _should be implemented_. Make sure the extension is available in `TusCapabilities`.

ylangisc (changes requested):

tus/src/main/java/ch/cyberduck/core/tus/TusUploadFeature.java
ylangisc: The spec says _should_ return 404, 410 or 403 but not _must_. Makes sense then.

tus/src/main/java/ch/cyberduck/core/tus/TusUploadFeature.java
dkocher: I see no benefit adding support for the upload creation extension.

ylangisc: Don't talk about the `creation-with-upload` extension. It's about the used `creation` extension here which is, as mentioned, optional and advertised in the capabilities if supported.

tus/src/main/java/ch/cyberduck/core/tus/TusUploadFeature.java
```diff
@@ -0,0 +1,250 @@
...
+            long remaining = status.getLength();
+            final String uploadUrl;
+            if(status.isAppend()) {
+                uploadUrl = preferences.getProperty(toUploadUrlPropertyKey(host, file));
+                if(log.isDebugEnabled()) {
+                    log.debug(String.format("Resume upload to %s for %s from offset %d", uploadUrl, file, status.getOffset()));
+                }
+            }
+            else {
+                if(!capabilities.extensions.contains(Extension.creation)) {
```
ylangisc, line 105: Move check to `Session#connect` to determine which `Upload` feature to take - based on version and creation extension. If the check is here only the client is unable to upload anything to a server with TUS support but no creation extension impl.

ylangisc (changes requested):

ylangisc (approved):

PR #15819, opened 2024-04-10
Add forNameOrDefault to ProtocolFactory

### LINKED ISSUE #15795: Crash with ArgumentNullException in System.Windows.Forms.CurrencyManager when loading Preferences
**Describe the bug**
When attempting to adjust preferences by going to the Preferences menu item, the app crashes. Therefore I can't access the latest snapshot builds. I've tried reverting to 8.8.0 and 8.7.3 and repairing the install of 8.8.1.41263, but it still crashes.

**To Reproduce**
1. Go to Preferences

**Expected behavior**
Preferences load

**Screenshots**
Cyberduck

 - OS: Windows 11
 - Version: 23H2 build 22631.3296

**Log Files**
The following is the crash report [638477617558948223.txt](https://github.com/iterate-ch/cyberduck/files/14855373/638477617558948223.txt)

AliveDevil:
Fixes #15795

### DIFF
--- a/core/src/main/java/ch/cyberduck/core/ProtocolFactory.java
+++ b/core/src/main/java/ch/cyberduck/core/ProtocolFactory.java
@@ -37,6 +37,7 @@ import java.util.Comparator;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.Predicate;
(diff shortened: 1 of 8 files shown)


core/src/main/java/ch/cyberduck/core/ProtocolFactory.java
```diff
@@ -219,32 +221,91 @@ public Protocol forName(final String identifier, final String provider) {
      * @param provider   Custom inherited protocol definition
      * @return Matching protocol or null if no match
      */
-    public Protocol forName(final List<Protocol> enabled, final String identifier, final String provider) {
-        final Protocol match =
```
AliveDevil, line 223: Removed Streams, because `orElse()` uses a parameter, and is not lazily evaluated.
Thus _all_ streams were evaluated _at once_, before returning.
Changed to range of checks, because `orElseGet` isn't readable.

core/src/main/java/ch/cyberduck/core/ProtocolFactory.java
```diff
@@ -219,32 +221,91 @@ public Protocol forName(final String identifier, final String provider) {
...
+            final String provider) {
+        final Stream<Protocol> stream = enabled.stream().sorted(new DeprecatedProtocolComparator());
+
+        // Exact match with hash code
+        Optional<Protocol> match = stream
+                .filter(protocol -> String.valueOf(protocol.hashCode()).equals(identifier))
+                .findFirst();
+        if (!match.isPresent()) {
+            // Matching vendor string for third party profiles
+            match = stream.filter(protocol -> StringUtils.equals(protocol.getIdentifier(), identifier)
```
dkocher, line 236: Stream has already been consumed.

core/src/main/java/ch/cyberduck/core/ProtocolFactory.java
```diff
@@ -219,32 +221,91 @@ public Protocol forName(final String identifier, final String provider) {
...
+                    && StringUtils.equals(protocol.getProvider(), provider)).findFirst();
+        }
+
+        if (!match.isPresent()) {
```
dkocher, line 240: Apply default formatting.

dkocher (changes requested):

core/src/main/java/ch/cyberduck/core/ProtocolFactory.java
AliveDevil: Done.

core/src/main/java/ch/cyberduck/core/ProtocolFactory.java
```diff
@@ -220,33 +222,101 @@ public Protocol forName(final String identifier, final String provider) {
      * @return Matching protocol or null if no match
      */
     public Protocol forName(final List<Protocol> enabled, final String identifier, final String provider) {
-        final Protocol match =
+        int state = 0; // Acts as findFirst-guard
+        Protocol match = null;
+        final Iterable<Protocol> protocols = enabled.stream().sorted(new DeprecatedProtocolComparator())::iterator;
+        for(final Protocol protocol : protocols) {
```
AliveDevil, line 228: As in https://github.com/iterate-ch/cyberduck/pull/15819/files#r1559353773, but:
As Streams cannot be reused, make one loop over all enabled protocols, and perform comparisons once.
Will only perform comparisons if applicable (state gated).

core/src/main/java/ch/cyberduck/core/ProtocolFactory.java
```diff
@@ -247,6 +274,38 @@ public Protocol forName(final List<Protocol> enabled, final String identifier, f
...
+     * @param provider   Custom inherited protocol definition
+     * @return Matching protocol or default if no match
+     */
+    public Protocol forNameOrDefault(final List<Protocol> enabled, final String identifier, final String provider) {
+        final String defaultIdentifier = preferences.getDefault("connection.protocol.default");
+        final Protocol match = forName(enabled, identifier, provider);
+        if(null != match || Objects.equals(defaultIdentifier, identifier)) {
+            return match;
+        }
+        return forName(enabled, identifier, null);
```
AliveDevil, line 306:
```suggestion
        return forName(enabled, identifier, null);
        return forName(enabled, defaultIdentifier, null);
```

AliveDevil, line 306:
```suggestion
        return forName(enabled, defaultIdentifier, null);
```

dkocher: Wir müssen noch Testabdeckung dafür haben.

AliveDevil: Done, forNameOrDefault is 100% covered.

dkocher (changes requested):
Sqash changes into single commit.

dkocher (approved):

core/src/main/java/ch/cyberduck/core/ProtocolFactory.java
```diff
@@ -220,24 +221,50 @@ public Protocol forName(final String identifier, final String provider) {
...
+                // Fallback for bug in 6.1
+                match = protocol;
+                state = 2;
+            }
+            else if(state < 1 && protocol.isBundled() && StringUtils.equals(protocol.getIdentifier(), identifier)) {
+                // Matching bundled first with identifier match
+                match = protocol;
+                state = 1;
+            }
+            else if(null == match && StringUtils.equals(protocol.getType().name(), identifier)) {
```
ylangisc, line 259: Necessary? `state` remains `0` and thus overridden in line 266 if there is no other match.

AliveDevil: It is, because the state=0 check should fire only if nothing in the loop ever matched.
Fixed in #726bfa8

dkocher: Add test coverage?

AliveDevil: Specifically:
The `null == match`-check in this loop is this part of the previous implementation:
```
enabled.stream().sorted(new DeprecatedProtocolComparator()).filter(protocol -> StringUtils.equals(protocol.getType().name(), identifier)).findFirst().orElse(null)
```
which is used later in
```
this.forScheme(enabled, identifier, …)
```
As the forScheme-fallback is only used after all previous streams did not return a match, this must be fetched later, and overwrite the type-name match.

ylangisc (approved):

dkocher (changes requested):
Squash commits.

dkocher (approved):

PR #15868, opened 2024-04-23
Only share context in subsequent logical requests

### LINKED ISSUE #15127: Unexpected 401 when using SharePoint WebDav NTLM authentication
With multiple connections opened for transferring items, there is a concurrency issue with the NTLM authentication state shared in the shared request context.

> IMPORTANT: Please note HTTP context implementation, even when thread safe, may not be used concurrently by multiple threads, as the context may contain thread unsafe attributes.

[HttpContext (Apache HttpComponents Core HTTP/1.1 5.2.2 API)](https://hc.apache.org/httpcomponents-core-5.2.x/5.2.2/httpcore5/apidocs/org/apache/hc/core5/http/protocol/HttpContext.html)

1. SardineImpl uses singular cached HttpContext for all requests[^1] 
2. Only one `DavClient extends SardineImpl` is used for all Dav requests[^2]
3. SardineImpl uses a request local HttpClientContext[^3]
4. But at the same time uses a broad-scoped AuthState-cache [^4] which is the underlying issue: One AuthState-cache for all request going out. This would need to be configurable or request local scoped, instead of global.

[^1]: https://github.com/iterate-ch/sardine/blob/f519c959582d9d461927100cc444ea996e209232/src/main/java/com/github/sardine/impl/SardineImpl.java#L169
[^2]: https://github.com/iterate-ch/cyberduck/blob/ebb964cd6761dfb286e44e931bde41072bd9d5a1/webdav/src/main/java/ch/cyberduck/core/dav/DAVSession.java#L107
[^3]: https://github.com/iterate-ch/sardine/blob/f519c959582d9d461927100cc444ea996e209232/src/main/java/com/github/sardine/impl/SardineImpl.java#L1055
[^4]: https://github.com/iterate-ch/sardine/blob/f519c959582d9d461927100cc444ea996e209232/src/main/java/com/github/sardine/impl/SardineImpl.java#L258-L262

dkocher:
Fix #15127. 

- [x] Ensure request context with authentication state is not shared between connections
- [ ] During the NTLM handshake HTTP requests must not contain a body but have `Content-Length: 0`. Otherwise Microsoft IIS will respond with `400 Bad Request` 
- [ ] When no connection is available from the pool for a `PUT` request we must ensure the entity is repeatable

### DIFF
--- a/core/src/main/java/ch/cyberduck/core/http/CustomHttpRequestExecutor.java
+++ b/core/src/main/java/ch/cyberduck/core/http/CustomHttpRequestExecutor.java
@@ -18,9 +18,12 @@ package ch.cyberduck.core.http;
  * feedback@cyberduck.io
  */
 
+import ch.cyberduck.core.Host;
 import ch.cyberduck.core.PreferencesUseragentProvider;
 import ch.cyberduck.core.TranscriptListener;
 import ch.cyberduck.core.UseragentProvider;
+import ch.cyberduck.core.exception.RetriableAccessDeniedException;
+import ch.cyberduck.core.preferences.HostPreferences;
 
 import org.apache.commons.lang3.StringUtils;
 import org.apache.http.Header;
@@ -29,20 +32,28 @@ import org.apache.http.HttpException;
 import org.apache.http.HttpHeaders;
 import org.apache.http.HttpRequest;
 import org.apache.http.HttpResponse;
+import org.apache.http.HttpStatus;
+import org.apache.http.client.HttpResponseException;
+import org.apache.http.client.methods.HttpPatch;
+import org.apache.http.client.methods.HttpPost;
+import org.apache.http.client.methods.HttpPut;
 import org.apache.http.message.BasicHeader;
 import org.apache.http.protocol.HttpContext;
 import org.apache.http.protocol.HttpRequestExecutor;
 
 import java.io.IOException;
+import java.util.Arrays;
 
-public class LoggingHttpRequestExecutor extends HttpRequestExecutor {
+public class CustomHttpRequestExecutor extends HttpRequestExecutor {
 
     private final UseragentProvider useragentProvider
             = new PreferencesUseragentProvider();
 
+    private final Host host;
     private final TranscriptListener listener;
 
-    public LoggingHttpRequestExecutor(final TranscriptListener listener) {
+    public CustomHttpRequestExecutor(final Host host, final TranscriptListener listener) {
+        this.host = host;
         this.listener = listener;
     }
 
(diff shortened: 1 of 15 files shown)


dkocher:
Will also need to address #2931.

core/src/main/java/ch/cyberduck/core/http/DelayedHttpEntity.java
```diff
@@ -77,7 +77,7 @@ public DelayedHttpEntity(final Thread parentThread, final CountDownLatch streamO
     private final Thread parentThread;
 
     public boolean isRepeatable() {
-        return !entityWritten;
+        return false;
```
ylangisc, line 80: Not sure if this affects other cases in a negative manner, i.e. https://github.com/iterate-ch/sardine/blob/master/src/main/java/com/github/sardine/impl/SardineImpl.java#L959

core/src/main/java/ch/cyberduck/core/http/CustomHttpRequestExecutor.java
```diff
@@ -77,8 +87,28 @@ protected HttpResponse doSendRequest(final HttpRequest request, final HttpClient
...
         final HttpResponse response = super.doSendRequest(request, conn, context);
         if(null != response) {
-            // response received as part of an expect-continue handshake
+            // Response received as part of an expect-continue handshake
             this.log(response);
+            if(new HostPreferences(host).getBoolean("request.unauthorized.ntlm.preflight")) {
+                // Unable to authenticate with NTLM for requests with non-zero entity
+                if(HttpStatus.SC_UNAUTHORIZED == response.getStatusLine().getStatusCode()) {
+                    if(response.containsHeader(HttpHeaders.WWW_AUTHENTICATE)) {
+                        switch(request.getRequestLine().getMethod()) {
```
ylangisc, line 96: Also add `POST` to make it more generic for non-WebDAV HTTP protocols supporting NTLM?

webdav/src/main/java/ch/cyberduck/core/dav/DAVSession.java
```diff
@@ -348,24 +372,34 @@ public <T> T _getFeature(final Class<T> type) {
...
+        public HttpCapabilities(final PreferencesReader preferences) {
+            this.expectcontinue = preferences.getBoolean("webdav.expect-continue");
+        }
+
+        public HttpCapabilities withExpectcontinue(final boolean expectcontinue) {
+            this.expectcontinue = expectcontinue;
+            return this;
+        }
+
+        public HttpCapabilities withIis(final boolean iis) {
```
ylangisc, line 391: Rename to `withIIS` to make it consistent with the classes such as `MicrosoftIISPersistentAuthResponseInterceptor`?

core/src/main/java/ch/cyberduck/core/http/DelayedHttpEntity.java
dkocher: Does not change semantics as the library will only query `isRepeatable` *after* invoking `writeTo` of the entity. Thus, the result is the same.

ylangisc: Also in the mentioned case above which is about `Expect: 100-continue` with servers no supporting it. Entity is not written yet, I guess.

dkocher: Retry handled in 84e2f5e5.

ylangisc (approved):

PR #15954, opened 2024-05-23
Preset domain and username from environment on Windows

### LINKED ISSUE #15953: Preset credentials from Windows when setting up WebDAV or SMB connection

dkocher:
Fix #15953.


ylangisc (commented):
Most use cases with WebDAV will not require the Windows user as username thus confusing.

dkocher:
> Most use cases with WebDAV will not require the Windows user as username thus confusing.

Should we remove it for WebDAV and only keep it for SMB?

ylangisc (approved):

PR #16033, opened 2024-06-17
Review transfer status and return value of features

dkocher:


ylangisc (commented):
Shouldn't `DAVTouchFeature` also just return the file as `DAVDirectoryFeature` now does? Both workers retrieve the attributes if necessary. Consistency...

dkocher:
> Shouldn't `DAVTouchFeature` also just return the file as `DAVDirectoryFeature` now does? Both workers retrieve the attributes if necessary. Consistency...

Missed in dbdf599a.

ylangisc (approved):

dkocher:
Will also review `Copy` and `Move` features for consistency.

ylangisc (approved):

PR #16046, opened 2024-06-19
Make sure not to manipulate Permission.EMPTY.

dkocher:


cryptomator/src/main/java/ch/cyberduck/core/cryptomator/CryptoVault.java
```diff
@@ -498,10 +498,11 @@ public Path decrypt(final Session<?> session, final Path file) throws Background
                         ciphertext, file.getParent().attributes().getDirectoryId().getBytes(StandardCharsets.UTF_8));
                 final PathAttributes attributes = new PathAttributes(file.attributes());
                 if(this.isDirectory(inflated)) {
-                    final Permission permission = attributes.getPermission();
+                    final Permission permission = new Permission(attributes.getPermission());
```
chenkins, line 501: Return new `Permission` in `or`/`and`/`not` instead? At least in case of `Permission. EMPTY`? Then, add test which shows that we "make sure not to manipulate `Permission.EMPTY`"?

ylangisc (approved):

PR #16074, opened 2024-06-25
Always chunk uploads. Max file size in non-resumable upload is limite…

dkocher:
…d to 2 GiB.


ylangisc (commented):
Introduce `StoregateThresholdWriteFeature` just like `EueThresholdWriteFeature`?

ylangisc (approved):

PR #16118, opened 2024-07-09
Migrate ikvmnet

### LINKED ISSUE #16119: Support arbitrary path lengths in Cyberduck
Currently the entire codebase is not long-path compatible:
- .NET Framework does not auto-prefix long-paths with `\\?\`
- Using File.IO-APIs in a `LongPathAware`-process doesn't work on `LongPathsEnabled=false`-systems
- legacy IKVM does not support automatic prefixing of long paths with `\\?\`
- legacy IKVM does not support long-path prefixed paths in nio i.e. `Paths.get(@"\\?\C:")`

Possible options:
Drop usage of `java.nio.file` in _all of_ Cyberduck (and all dependencies) - does not handle long paths issues in System.IO
Add polyfill for System.IO, which automatically prefixes everything with `\\?\` as required - does not handle long path issues in `java.nio.file`.

With integration of [ikvmnet](https://github.com/ikvmnet/ikvm) we detach from the legacy IKVM implementation, which reimplemented huge parts of the JDK using .NET technology, with shipping rebuilds of the JDK native code, which restores real JDK behavior (i.e. everything the JDK has built for e.g. java.nio.file does work now). This includes automatic prefixing of long-paths with `\\?\` and support of long paths in `java.nio.file`.

Until migrating to .NET 8 (or later) an Api analyzer is added to Cyberduck, which bans all of System.IO.

AliveDevil:
Preparation work to get ikvmnet integrated as soon as a nuget package for 8u312 is available.

Resolves #16119


dkocher:
Superseded by #16157.

PR #16144, opened 2024-07-16
Fix DeepBox integration tests.

chenkins:
Fixes `testRewriteOverridenFileByRename` and `testSyncConflictDirectoryCreatedLocalAndRemoteIgnoreConflict`.

### DIFF
--- a/deepbox/src/main/java/ch/cyberduck/core/deepbox/DeepboxDirectoryFeature.java
+++ b/deepbox/src/main/java/ch/cyberduck/core/deepbox/DeepboxDirectoryFeature.java
@@ -25,6 +25,7 @@ import ch.cyberduck.core.deepbox.io.swagger.client.model.Folder;
 import ch.cyberduck.core.deepbox.io.swagger.client.model.FolderAdded;
 import ch.cyberduck.core.exception.AccessDeniedException;
 import ch.cyberduck.core.exception.BackgroundException;
+import ch.cyberduck.core.exception.ConflictException;
 import ch.cyberduck.core.features.Directory;
 import ch.cyberduck.core.features.Write;
 import ch.cyberduck.core.transfer.TransferStatus;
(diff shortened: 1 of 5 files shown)


deepbox/src/main/java/ch/cyberduck/core/deepbox/DeepboxMoveFeature.java
```diff
@@ -68,7 +68,7 @@ public Path move(final Path file, final Path renamed, final TransferStatus statu
             new CoreRestControllerApi(session.getClient()).updateNode(nodeUpdate, sourceId);
             fileid.cache(file, null);
             fileid.cache(renamed, sourceId);
-            return renamed;
+            return renamed.withAttributes(new DeepboxAttributesFinderFeature(session, fileid).find(renamed));
```
ylangisc, line 71: Copy original file attributes:
```
return renamed.withAttributes(file.attributes());
```

deepbox/src/main/java/ch/cyberduck/core/deepbox/DeepboxDirectoryFeature.java
```diff
@@ -52,6 +53,9 @@ public DeepboxDirectoryFeature(final DeepboxSession session, final DeepboxIdProv
     @Override
     public Path mkdir(final Path folder, final TransferStatus status) throws BackgroundException {
         try {
+            if(new DeepboxFindFeature(session, fileid).find(folder)) {
```
ylangisc, line 56: Use `status.isExists()` and trash folder when existing. Refer to `DriveMoveFeature#move`.

chenkins: Copied from `DriveDirectoryFeature`.

dkocher: `CreateDirectoryWorker` and `TouchWorker` do not determine if a target might already exist. Thus we cannot rely on the `exists` flag in this case but expect implementations to throw failure on conflict. This was introduced in #14168.

dkocher (changes requested):
Add test coverage.

dkocher:
We should also fix `DeepboxCopyFeature` to return attributes of target with 

```
return target.withAttributes(new DeepboxAttributesFinderFeature(session, fileid).toAttributes(copied))
```

dkocher (approved):

PR #16168, opened 2024-07-24
Notify TransfersWindow to reveal an item

### LINKED ISSUE #16073: Transfer window opens completely scrolled to the top
**Describe the bug**
Transfer window opens completely scrolled to the top so that only the old transfers are visible.

**Expected behavior**
Open Transfer window with focus on the current transfer.

 - OS: Windows
 - Cyberduck Version 9

AliveDevil:
Fixes #16073

### DIFF
--- a/windows/src/main/csharp/ch/cyberduck/ui/ViewModels/TransfersViewModel.cs
+++ b/windows/src/main/csharp/ch/cyberduck/ui/ViewModels/TransfersViewModel.cs
@@ -26,6 +26,7 @@ using DynamicData;
 using DynamicData.Aggregation;
 using DynamicData.Binding;
 using org.apache.logging.log4j;
+using ReactiveUI;
 using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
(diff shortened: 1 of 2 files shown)


windows/src/main/csharp/ch/cyberduck/ui/ViewModels/TransfersViewModel.cs
```diff
@@ -233,8 +234,12 @@ public bool ToolbarSeparatorRight
         var transfersCache = store.Transfers.Connect().ObserveOnDispatcher()
             .Transform(m => new TransferViewModel(controller, m, locale)).DisposeMany()
             .Bind(out transfers).AsObservableCache().DisposeWith(subscriptions);
-        subscriptions.Add(transfersCache.Connect().Subscribe(OnTransfersChanged));
         subscriptions.Add(transfersCache.CountChanged.Subscribe(OnTransfersCountChanged));
+        subscriptions.Add(Observable.FromEvent<Transfer>(
+            h => controller.RevealTransfer += h,
+            h => controller.RevealTransfer -= h)
+            .Select(transfersCache.WatchValue).Switch()
```
AliveDevil, line 241: If transfersCache doesn't have Transfer cached, with Lookup would return `Optional.None`.
This guarantees that a valid value is passed on.

windows/src/main/csharp/ch/cyberduck/ui/ViewModels/TransfersViewModel.cs
```diff
@@ -233,8 +234,12 @@ public bool ToolbarSeparatorRight
...
+            .ObserveOnDispatcher().Subscribe(OnRevealTransfer));
```
AliveDevil, line 242: UI Thread synchronization required

windows/src/main/csharp/ch/cyberduck/ui/controller/TransferController.cs
```diff
@@ -318,6 +320,8 @@ void TransferListener.transferDidStart(Transfer t)
             }
 
             progress.Refresh(null);
+
+            RevealTransfer?.Invoke(t);
```
AliveDevil, line 324: Each started transfer should allow the UI to select that transfer.

ylangisc (approved):

PR #16175, opened 2024-07-24
Fix passing bad remote filenames to IconProvider

### LINKED ISSUE #16174: Invalid path characters on filenames quit application
Regression from #15912

```
Top-level Exception
Type:        System.ArgumentException
Message:     Illegal character in path.
Source:      mscorlib
Stack Trace: bei System.IO.Path.CheckInvalidPathChars(String path, Boolean checkAdditional)
   at System.IO.Path.IsPathRooted(String path)
   at Ch.Cyberduck.Core.Refresh.Services.IconProvider`1.GetFileIcon(String filename, Boolean isFolder, Boolean large, Boolean isExecutable)
   at Ch.Cyberduck.Core.Refresh.Services.IconProvider`1.GetPath(Path path, Int32 size)
   at BrightIdeasSoftware.ObjectListView.MakeSubItem(Object rowObject, OLVColumn column)
   at BrightIdeasSoftware.ObjectListView.FillInValues(OLVListItem lvi, Object rowObject)
   at BrightIdeasSoftware.VirtualObjectListView.MakeListViewItem(Int32 itemIndex)

```

AliveDevil:
Fixes #16174

### DIFF
--- a/core/native/refresh/src/main/csharp/Services/IconProvider.Path.cs
+++ b/core/native/refresh/src/main/csharp/Services/IconProvider.Path.cs
@@ -7,8 +7,8 @@ namespace Ch.Cyberduck.Core.Refresh.Services
     {
         public T GetPath(Path path, int size)
         {
-            string key = "path:" + (path.isDirectory() ? "folder" : path.getExtension());
-
+            var fileIconKey = path.getExtension();
+            string key = "path:" + (path.isDirectory() ? "folder" : fileIconKey);
             Func<(string, Func<int, T>)> overlayFactory = default;
             if (path.getType().contains(AbstractPath.Type.decrypted))
             {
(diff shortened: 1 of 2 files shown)


core/native/refresh/src/main/csharp/Services/IconProvider.Path.cs
```diff
@@ -53,7 +48,7 @@ public T GetPath(Path path, int size)
                 return image;
             }
 
-            var baseImage = GetFileIcon(path.getExtension(), path.isDirectory(), size >= 32, false);
+            var baseImage = GetFileIcon($".{fileIconKey}", path.isDirectory(), size >= 32, false);
```
ylangisc, line 51: Correct? Before it was the extension only. Now with a leading `.`.

AliveDevil: Yes, path.getExtension is `ext` instead of `.ext` and SHGetFileInfo doesn't work when leading `.` is missing.

ylangisc (approved):

PR #16233, opened 2024-08-15
Disable custom versioning.

dkocher:


deepbox/src/main/java/ch/cyberduck/core/deepbox/DeepboxProtocol.java
```diff
@@ -63,4 +63,9 @@ public boolean isUsernameConfigurable() {
     public DirectoryTimestamp getDirectoryTimestamp() {
         return DirectoryTimestamp.explicit;
     }
+
+    @Override
+    public VersioningMode getVersioningMode() {
+        return VersioningMode.storage;
```
chenkins, line 69: Confusing, backend only allows replacing file content of a `nodeId`, but the different versions have no referer/ID, there is only "latest": https://api.deepbox.swiss/swagger-ui/index.html#/revision-rest-controller/replaceCurrentRevision

Add comment/test coverage documenting the change.

chenkins (changes requested):

deepbox/src/main/java/ch/cyberduck/core/deepbox/DeepboxProtocol.java
dkocher: Versioning is handled by server and might be exposed via API in the future. We do not want to support our custom versioning scheme for this protocol.

PR #16256, opened 2024-08-26
Skip reachability check when only proxy jump host connection is possi…

dkocher:
…ble.


core/src/main/java/ch/cyberduck/core/diagnostics/ReachabilityFactory.java
```diff
@@ -50,6 +55,12 @@ public void test(final Host bookmark) throws BackgroundException {
                 case http:
                     new ChainedReachability(monitor, new ResolverReachability(), new HttpReachability()).test(bookmark);
                     break;
+                case sftp:
+                    if(null != JumpHostConfiguratorFactory.get(bookmark.getProtocol()).getJumphost(bookmark.getHostname())) {
```
AliveDevil, line 59: Couldn't this test for reachability of the jump host instead?

AliveDevil (approved):

ylangisc (approved):

PR #16282, opened 2024-09-03
Fix #16281.

### LINKED ISSUE #16281: SSH non-default port setting not being read from .ssh/config file
Hi,

At least on my system, Cyberduck isn't reading the `Port` parameter from `~/.ssh/config`. While it does read the host's IP, `IdentityFile` etc…, the port is always 22 after updating other data in the config file.

Steps to reproduce:
- Create bookmark using a host alias from SSH's config file.
- Update said config file, changing both IP and Port (e.g. when using cloud GPUs)
- Edit bookmark: you'll see that the displayed IP is the new one, but Port: has the same value it had before changing the config file.

Thanks!

dkocher:


osx/src/main/java/ch/cyberduck/ui/cocoa/controller/BookmarkController.java
```diff
@@ -216,11 +216,7 @@ public void protocolSelectionChanged(final NSPopUpButton sender) {
                 bookmark.setDefaultPath(selected.getDefaultPath());
             }
             bookmark.setProtocol(selected);
-            final int port = HostnameConfiguratorFactory.get(selected).getPort(bookmark.getHostname());
-            if(port != -1) {
-                // External configuration found
-                bookmark.setPort(port);
-            }
+            bookmark.setPort(HostnameConfiguratorFactory.get(selected).getPort(bookmark.getHostname()));
```
ylangisc, line 219: Check if `-1` not necessary anymore? Without check the default port is always overridden even if there is no external port set.

dkocher: Changing the protocol will now set the default port regardless of any previous entry.

ylangisc: Don't get it. In line 200 the default port is set but overridden here with `-1` unless it's SSH with an alias in `.ssh/config` and custom port.

dkocher: Setting with parameter `-1` will reset to the default port of the current protocol. Will remove the redundant setter.

ylangisc (approved):

PR #13030, opened 2022-03-31
Edit capability for Store-apps

AliveDevil:
… and ProgIDs, and everything else. Doesn't rely on registry lookups, command handling etc.
Uses basic standard shell functions for achieving ths.

Closes #12802

PR #16393, opened 2024-10-01
Remove "Choose…" button from preferences default editor selection

### LINKED ISSUE #16387: On Windows 11 inability to select Notepad++ as default editor
When I try to change preferences editor no file dialog to select my main editor Notepad++

Other applications are listed without needing to browse to.

But if you need to select other there's no file dialog to let you select it.

**To Reproduce**
1. Go to 'Preferences'
2. Click on 'Editor'
3. Select 'Choose'
4. Notice Nothing Happens then
5. Close 'Preferences'
6. In the main bar Next to editor button, Dropdown triangle
7. Select 'Choose'
8. File dialog option says 'Select an App to open .html' file
9. There's only the option to use this one time

**Expected behavior**
I would expect the app the see from my system that .txt files are usually opened with notepad++ and offer this as one of the options.
If not that, at minimum, I would expect the file dialog to be invoked when I click step 3 'Choose' the .exe of my editor and then to save and persist in memory.

**Screenshots**
File Dialog Just one time option not save as default
bug select editor notepad++_not-prepopulated
bug select editor notepad++_not-prepopulated+Choose2
bug select editor notepad++_not-prepopulated+Choose
bug select editor

Initiate - File Dialog Just one time option not save as default

Windows 11. 22631.4169

**Log Files**
[cyberduck.log](https://github.com/user-attachments/files/17179640/cyberduck.log)

**Additional context**
I usually have a bit of French and English as I live in Canada.

AliveDevil:
This doesn't have any effect anymore (#16387)
Missed in #13030

### DIFF
--- a/windows/src/main/csharp/ch/cyberduck/ui/controller/PreferencesController.cs
+++ b/windows/src/main/csharp/ch/cyberduck/ui/controller/PreferencesController.cs
@@ -1408,8 +1408,7 @@ namespace Ch.Cyberduck.Ui.Controller
                             defaultEditor.getName()));
                 }
             }
-            editors.Add(new KeyValueIconTriple<Application, string>(Application.notfound,
-                LocaleFactory.localizedString("Choose") + "…", String.Empty));
+
             View.PopulateEditors(editors);
             if (defaultEditor != null)
             {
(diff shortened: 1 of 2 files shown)


ylangisc:
Remove the corresponding part in [PreferencesForm](https://github.com/iterate-ch/cyberduck/blob/master/windows/src/main/csharp/ch/cyberduck/ui/winforms/PreferencesForm.cs#L1509-L1532)?

ylangisc (approved):

PR #16440, opened 2024-10-17
Set progress when concatenating files.

### LINKED ISSUE #13610: No progress when concatenating downloaded segments for large download after transfer is complete
**Describe the bug**

I am using Backblaze S2, however I suspect that this issue will happen with all storage providers that support multipart downloads.   In my circumstance, I was downloading a 27GB file from Backblaze, which eventually hit 100% completion, however he transfers window still shows it as "Downloading" for several minutes while the speed drops slowly, before it says complete.   While it is stuck at 100% downloading, i can see the file in the downloads directory on my computer increasing in size until it hits the correct file size, at which point the transfers window shows the download as complete.   I suspect that as I have the "Segmented downloads with multiple connections per file" setting enabled, cyberduck is reconstructing the file, however the status on the transfers window doesn't change to indicate this, which could lead to confusion or people thinking downloads are not working correctly.

**To Reproduce**

1. Ensure Segmented download with multiple connections per file is enabled
2. Download a file tens of gigabytes or more in size
3. Wait until the download hits 100%

**Expected behavior**
If the issue is what i suspect it to be, i would expect the transfers window to change from Downloading 100% to something along the lines of either Reconstructing or Building InsertFileNameHere, which a percentage indicator showing its progress.

Windows 11 21H2 (22000.795)

dkocher:
Fix #13610.


core/src/main/java/ch/cyberduck/core/transfer/download/AbstractDownloadFilter.java
```diff
@@ -282,6 +282,8 @@ public void complete(final Path file, final Local local,
                 if(local.exists()) {
                     local.delete();
                 }
+                listener.message(MessageFormat.format(LocaleFactory.localizedString("Finalize {0}", "Status"),
```
AliveDevil, line 285: Can we make this more userfriendly, and include ProgressReporting capabilities to the target-file WriteChannel?

AliveDevil:
https://github.com/user-attachments/assets/b9ef481e-d433-4f04-a4e5-c95eaeeb03f2

Need to assess whether this behavior is tolerable.

dkocher:
> Need to assess whether this behavior is tolerable.

It would be preferrable if the progress bar would change to indeterminate spinning while finalizing.

AliveDevil:
So, while there is Progress in progress, disable Show/Open and do not rely on `Transfer.completed`?

AliveDevil (changes requested):
Draft: Add Indeterminate state to macOS interface.

windows/src/main/csharp/ch/cyberduck/ui/ViewModels/TransfersViewModel.cs
```diff
@@ -608,6 +615,7 @@ private void ValidateCommands()
         StopCommand.NotifyCanExecuteChanged();
         TrashCommand.NotifyCanExecuteChanged();
 
+        // TODO Reduce triggering this a billion times a second
```
dkocher, line 618: @AliveDevil Resolve TODO

AliveDevil: Not now. Would require reevaluating how all commands/notifications are handled in this Window. Future improvement.

ylangisc (approved):

PR #16471, opened 2024-10-31
For stateful connections (FTP) limit to a maximum of 5 connections lo…

dkocher:
…wering to 1 on failures.


ylangisc:
Would make that protocol dependant (`connection.pool.maxtotal.*`) as we have with `queue.connections.limit.*`.

dkocher:
@ylangisc This pool is only used for stateful protocols.

ylangisc (approved):

PR #16473, opened 2024-10-31
Fix #16377

### LINKED ISSUE #16377: Unable to connect to Cloudflare R2 owing to lack of ability to change region
**Describe the bug**
I'm trying to connect to an R2 bucket with Cyberduck using the Cloudflare R2 profile. However, I'm greeted with the following error:

> The region name 'ap-southeast-1' is not valid. Must be one of: wnam, enam, weur, eeur, apac, auto. Please contact your web hosting service provider for assistance.

When I open the Cyberduck's settings (Settings > S3 > Default Bucket Location), I can change the region, and I can confirm the change is reflected, however the options `enam, weur, eeur, apac, auto` are not available.

**To Reproduce**
1. Install [Cloudflare R2 profile](https://docs.cyberduck.io/protocols/s3/cloudflare/)
2. Use Cloudflare R2 bucket credentials and add bucket name as path
3. Attempt to establish connection
4. The above error is shown

**Expected behavior**

 - OS: MacOS 15.0 (24A335)
 - Version 9.0.2 (42108)
 
**Additional context**
When connecting using boto3, the same error is shown when using "auto", "ap-southeast", or any other regions other than `enam, weur, eeur, apac, auto`. In my case, when I used `apac` instead of `ap-southeast`, the error disappeared. So I believe the ability to change the region in the client will fix the issue.

dkocher:


s3/src/main/java/ch/cyberduck/core/s3/S3Session.java
```diff
@@ -335,25 +342,45 @@ public void login(final LoginCallback prompt, final CancelCallback cancel) throw
...
+        else {
+            final Path home = new DelegatingHomeFeature(new DefaultPathHomeFeature(host)).find();
+            if(home.isRoot()) {
+                if(log.isDebugEnabled()) {
+                    log.debug(String.format("Skip querying region for %s", home));
+                }
+                new S3ListService(this, acl).list(home, new DisabledListProgressListener());
+            }
+            else {
+                final Location.Name location = new S3LocationFeature(this, regions).getLocation(home);
```
ylangisc, line 373: Why can the exception handling for `AccessDeniedException | InteroperabilityException` be omitted? What happens in this case now?

dkocher: See commit c78789d7a3ec2f468ff533dbc6c3e82fb35ec5cd

ylangisc: At least the Spectra tests are red - not clear if this is related to this specific change or another one in this PR.

dkocher: Spectra integration tests are ✅

ylangisc (approved):

PR #16480, opened 2024-11-03
Review log formatting to use parameterized messages.

dkocher:


ylangisc (commented):
`isDebugEnabled()` checks can be omitted now as no evaluation happens this way.

core/src/main/java/ch/cyberduck/core/threading/DefaultFailureDiagnostics.java
```diff
@@ -49,9 +49,7 @@ public final class DefaultFailureDiagnostics implements FailureDiagnostics<Backg
 
     @Override
     public Type determine(final BackgroundException failure) {
-        if(log.isDebugEnabled()) {
-            log.debug(String.format("Determine cause for failure %s", failure));
-        }
+        log.debug("Determine cause for failure {}", failure.getMessage());
```
ylangisc, line 52: NPE for `null` failure.

ylangisc (changes requested):

core/src/main/java/ch/cyberduck/core/threading/DefaultFailureDiagnostics.java
dkocher: Argument is  not expected to be `null`.

ylangisc: https://github.com/iterate-ch/cyberduck/blob/1034b879dd69a231c22851d2eeac438c9ec7a856/core/src/test/java/ch/cyberduck/core/threading/DefaultFailureDiagnosticsTest.java#L34

ylangisc (approved):

PR #16555, opened 2024-11-20
Include resource in If: header for lock tokens

dkocher:


webdav/src/main/java/ch/cyberduck/core/dav/DAVMoveFeature.java
```diff
@@ -49,18 +50,18 @@ public DAVMoveFeature(final DAVSession session) {
     public Path move(final Path file, final Path renamed, final TransferStatus status, final Delete.Callback callback, final ConnectionCallback connectionCallback) throws BackgroundException {
         try {
             final String target = new DefaultUrlProvider(session.getHost()).toUrl(renamed).find(DescriptiveUrl.Type.provider).getUrl();
-            if(session.getFeature(Lock.class) != null && status.getLockId() != null) {
+            if(status.getLockId() != null && session.getFeature(Lock.class) != null) {
                 // Indicate that the client has knowledge of that state token
                 session.getClient().move(new DAVPathEncoder().encode(file), file.isDirectory() ? String.format("%s/", target) : target,
                         status.isExists(),
-                        Collections.singletonMap(HttpHeaders.IF, String.format("(<%s>)", status.getLockId())));
+                        Collections.singletonMap(HttpHeaders.IF, String.format("<%s>(<%s>)", new DAVPathEncoder().encode(file), status.getLockId())));
```
ylangisc, line 57: According to spec there must be a space in between → `<%s> (<%s>)`

PR #16599, opened 2024-11-29
Cache passphrase to avoid expensive unlock operation.

ylangisc:


dracoon/src/main/java/ch/cyberduck/core/sds/SDSSession.java
```diff
@@ -136,6 +129,14 @@ public class SDSSession extends HttpSession<SDSApiClient> {
     private final ExpiringObjectHolder<SoftwareVersionData> softwareVersion
             = new ExpiringObjectHolder<>(preferences.getLong("sds.useracount.ttl"));
 
+    private final LRUCache<UserKeyPair, Credentials> keyPairPassphrases
```
dkocher, line 132: `UserKeyPair` does not implement `equals/hashCode`.

dkocher (changes requested):

dkocher (approved):

PR #16724, opened 2024-12-26
Always sort result set from DNS resolver

### LINKED ISSUE #16723: No fallback to IPv4 adress when IPv6 connect attempt fails with Network is unreachable: connect
Discussed in #16703.

The issue is that with #16639 IPv6 address may be preferred over IPv4 and no attempt is made to connect using IPv4 if an IPv6 address causes `java.net.SocketException) Network is unreachable: connect`.

dkocher:


core/src/main/java/ch/cyberduck/core/Resolver.java
```diff
@@ -67,20 +68,25 @@ public Resolver(final boolean preferIPv6) {
...
                     final InetAddress[] allByName = InetAddress.getAllByName(hostname);
-                    resolved.set(Arrays.stream(allByName).collect(Collectors.toSet()));
-                    if(preferIPv6) {
-                        final Set<InetAddress> filtered = Arrays.stream(allByName).filter(inetAddress -> inetAddress instanceof Inet6Address).collect(Collectors.toSet());
-                        if(!filtered.isEmpty()) {
-                            resolved.set(filtered);
+                    resolved.set(Arrays.stream(allByName).sorted(new Comparator<InetAddress>() {
+                        @Override
+                        public int compare(final InetAddress o1, final InetAddress o2) {
+                            if(o1 instanceof Inet6Address) {
```
ylangisc, line 81: Order of IPv6 addresses as in `allByName` is reversed when `preferIPv6` is `true`. Should be `o1 instanceof Inet6Address && o2 instanceof Inet4Address`. Same for the following test.

ylangisc (approved):

PR #16762, opened 2025-01-09
Make sure required feature is available regardless of connection state.

dkocher:


ftp/src/main/java/ch/cyberduck/core/ftp/FTPSession.java
```diff
@@ -304,7 +297,7 @@ public <T> T _getFeature(final Class<T> type) {
             return (T) new FTPDeleteFeature(this);
         }
         if(type == Read.class) {
-            return (T) read;
+            return (T) read;//todo null
```
ylangisc, line 300: `todo`?

ylangisc (approved):

PR #17001, opened 2025-03-31
Use existing methods of fetching application icons

### LINKED ISSUE #17000: Crash when trying to edit Preferences
**Describe the bug**
Try to edit preferences with a newly installed Cyberduck (Version 9.1.3) causes it to crash.

**To Reproduce**
Steps to reproduce the behaviour:
1. On a PC running Windows, that has previously never had Cyberduck installed.
2. Install the latest version of Cyberduck (Version 9.1.3). 
3. Try to edit preferences, using the keys "Ctrl+," or by selecting Preferences in the Edit menu.
4. Cyberduck crashes with the "Do you want to report the last crash dialog"

**Expected behaviour**
I expect to be able to edit Preferences.

**Screenshots**
Image

 - OS: Microsoft Windows 11
 - Version:  24H2 (OS Build 26100.3264)

**Log Files**
[638790108694596355.txt](https://github.com/user-attachments/files/19532667/638790108694596355.txt)

**Additional context**
- I don't think that a default preferences file was provided when Cyberduck was installed.
- Cyberduck does not appear to be able to edit preferences if there is no preferences file installed.
- Functional bookmarks can be added and Cyberduck does appear to be functional but Preferences cannot be edited.

AliveDevil:
Fixes #17000

### DIFF
--- a/windows/src/main/csharp/ch/cyberduck/ui/winforms/PreferencesForm.cs
+++ b/windows/src/main/csharp/ch/cyberduck/ui/winforms/PreferencesForm.cs
@@ -884,13 +884,13 @@ namespace Ch.Cyberduck.Ui.Winforms
             editorComboBox.IconMember = "IconKey";
             editorComboBox.DataSource = editors;
 
-            ImageList imageList = new ImageList();
+            ImageList imageList = new();
             foreach (KeyValueIconTriple<Application, string> triple in editors)
             {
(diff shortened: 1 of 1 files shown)


AliveDevil:
Without this patch:
image

With this patch:
image

dkocher (approved):

PR #17026, opened 2025-04-09
Unlinking file is allowed regardless of write permission.

dkocher:


ylangisc (commented):
Check if type is symlink instead of removing the whole check?

PR #17086, opened 2025-05-06
Rate-limit/Synchronize TaskDialog creation to one at a time

AliveDevil:
This resolves issues when two or more dialogs are opened simultaneously, that one or more of these will fail, with either 0x800403E9 (maybe Stack Overflow, but facility 4 (Interface) could be anything, basically an opaque error code), or some "Unknown" error.

The specific conditions when this appears are unknown, though I've been successful when trying to create two or more dialogs after application startup, before any other dialog has yet opened.

I.e. following won't fail:
- 1 dialog, then 10 dialogs
- 1 dialog, 1 dialog, 1 dialog, with 
- 2 dialogs - one failed, another 10 dialogs work.

These will fail, however:
- 2 dialogs - one, guaranteed
- 10 dialogs - up to 2, I've seen fail

### DIFF
--- a/core/src/main/csharp/ch/cyberduck/core/taskdialog/TaskDialog.cs
+++ b/core/src/main/csharp/ch/cyberduck/core/taskdialog/TaskDialog.cs
@@ -21,6 +21,7 @@ using System.Buffers;
 using System.Collections.Generic;
 using System.Drawing;
 using System.Runtime.InteropServices;
+using System.Threading;
 using Windows.Win32.Foundation;
 using Windows.Win32.UI.Controls;
 using Windows.Win32.UI.WindowsAndMessaging;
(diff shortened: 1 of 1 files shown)


core/src/main/csharp/ch/cyberduck/core/taskdialog/TaskDialog.cs
```diff
@@ -309,19 +313,24 @@ unsafe static void CopyButtons(ref TASKDIALOG_BUTTON target, List<TASKDIALOG_BUT
                 handle = GCHandle.Alloc(this, GCHandleType.Normal);
                 copy.lpCallbackData = GCHandle.ToIntPtr(handle);
 
+                syncLockAcquired = syncEvent.WaitOne(250);
```
AliveDevil, line 316: Wait for 250ms before continuing regardless.
Whatever failed/hanged shouldn't block other threads/dialogs.
From my testing, this timeout hasn't been an issue.

ylangisc (approved):

PR #17126, opened 2025-05-25
Release Jobs

### LINKED ISSUE #14998: Migrate all jobs from Jenkins CI to GitHub Actions
Convert existing jobs in Jenkins CI for development, test and deployment to GitHub Actions.

# Cyberduck 

### Unit Tests

- [x] [^1] 

[^1]: https://github.com/iterate-ch/cyberduck/actions/workflows/build.yml

### Deployment Jobs for macOS, Windows

Upload releases to S3/Rackspace from tag and create release in Github.

- [x] Snapshot Builds
- [x] Release Builds
- [x] Publish update feed
- [x] Publish Maven artifacts

### Windows

Builds for Windows must additionally be codesigned with custom certificate

- [x] Codesign builds with codesigning certificate 
- [ ] Publish build to Windows Store
- [x] Publish to Chocolatey Package Manager

### macOS

Builds for macOS must additionally be codesigned with certificate provided by Apple and [notarized](https://developer.apple.com/documentation/security/notarizing_macos_software_before_distribution).

- [x] Sign update feed for Sparkle
- [x] Codesign builds with codesigning certificate
- [x] Notarize builds from within build job or custom build action
- [x] Additional build with separate configuration for Mac App Store
- [x] Publish build to Mac App Store

## Integration Tests

- [x] Run integration tests [^2] 

[^2]: https://github.com/iterate-ch/cyberduck/actions/workflows/ci.yml

## Localization

- [x] Integrate localization updates from Transifex with pull request

# Cyberduck CLI (`duck`)

## Build Jobs

Builds for macOS, Windows and Linux (RPM, DEB).

- [ ] ~~Github Runner für duck-release-arm32~~
- [x] Github Runner für duck-release-aarch64
- [ ] Publish to Homebrew Package Manager

dkocher:
Resolve #14998.


ylangisc:
## Review — Release Jobs

**Note upfront:** the deploy workflows never ran on this PR (they only trigger on tags/master/dispatch), so the green checks don't cover any of the changed release plumbing.

### Critical

1. **Invalid expression in `deploy.yml`:** `runs-on: ${{ inputs['platform']-self-hosted }}` — GitHub expressions have no `-` operator/concatenation, so the workflow fails validation on every call. Presumably intended: `runs-on: ${{ inputs.platform }}`, consistent with `runs-on: linux` (publish job) and `runs-on: macos` (appstore.yml).

2. **`publish` runs after a failed build:** `if: ${{ inputs['publish'] && !cancelled() }}` drops the implicit `success()` check, and `!cancelled()` is true on `failure()`. Combined with `continue-on-error: true` on the download steps, a failed release build would silently publish a partial/empty artifact set to S3/Rackspace/SFTP. Drop `!cancelled()` or use `needs.build.result == 'success'`.

3. **`archive: false` breaks the update feed:** With `upload-artifact@v7` and `archive: false`, only a *single* file may be uploaded and the *file name becomes the artifact name — the `name` input is ignored*. The "Archive Update Feed" steps upload `changelog.*` (at least `changelog.rss` + `changelog.html`) under `name: cyberduck-<platform>-changelog`: multiple files → the step fails; even with a single file the artifact would be named `changelog.rss`, so the publish job's download by `pattern: cyberduck-<platform>-changelog` matches nothing — masked by `continue-on-error`. Net effect: the feed is never published and the workflow stays green. The other unnamed archive steps work with this semantic (filename = artifact name, which the `{*.zip,*.pkg,…}` download patterns rely on), but each depends on its glob matching exactly one file.

4. **Inconsistent `sparkle-feed` sentinel for releases:** `release.yml` passes `sparkle-feed: release`, but (a) the feed then uploads to `version.cyberduck.io/macos/release/changelog.rss` while the client default in this same PR is changed to `…/macos/changelog.rss`; (b) the CLI publish steps (`cdn.duck.sh`, `dist.duck.sh`) are gated on `sparkle-feed == ''` and therefore never run — no caller passes `''` with `publish: true`, so the duck CLI would no longer be deployed on releases. The S3/Rackspace fallback expressions already treat `''` as release. Either pass `''` from `release.yml` (and check what an explicitly empty `-Dsparkle.feed=` means vs. unset in the build) or gate on `== 'release'` and special-case the feed path.

### Windows code signing

- The default CNG path (`signtool sign … /f setup/cert/certificate.crt /csp … /kc …`) requires the SafeNet CSP installed, the token present, and both `SIGNTOOL_CSP`/`SIGNTOOL_KC` secrets configured (`certificate.crt` is checked in ✓). Classic Jenkins→GHA pitfall: if the runner runs as a Windows *service* rather than in the interactive session with the cached token PIN, signtool hangs on the PIN prompt — unless the PIN is embedded via the decorated KC syntax (`[{{PIN}}]=container`) in `SIGNTOOL_KC`. Worth verifying with a manual dispatch before the first tag build.
- **No fail-fast:** if the secrets are missing, Ant's `expandproperties` leaves the literal `${signtool.csp}` in the generated `Version.props` and signtool fails cryptically. An early workflow check (or `<fail unless=…>` when signing isn't skipped) would be more robust.
- **No signing-method selection:** `Cyberduck.Sign.targets` only knows `cng` and the certificate-store fallback with a hardcoded `/n "iterate GmbH"`; the reusable `deploy.yml` has no input to select method or certificate. Builds that need a different signing setup (e.g. white-label) would only work via Maven profiles overriding `signtool`/`codesign.certificate` (Maven properties do win over Ant's `<property>`). Consider adding a `signtool` input/env now — analogous to the `NOTARIZATION_TEAMID` un-hardcoding already in this PR.

### Further points

- **Maven artifacts deployed 4×:** `mvn deploy` runs in all four matrix legs, deploying the Java artifacts to maven.cyberduck.io each time — later legs fail if the release repo forbids redeploy; at minimum it's a race. Restrict artifact deployment to one leg.
- **Overlapping publish jobs:** artifacts are shared per workflow run; the four parallel publish jobs all download with the same unscoped patterns and concurrently re-upload the same files. `--existing overwrite` mostly saves it, but a platform-scoped selection would be cleaner.
- **Self-hosted runner hygiene:** the composite action *replaces* the user keychain search list (`security list-keychains -d user -s` — login.keychain is dropped) and never cleans up the keychain or `certificate.p12`; the Sparkle private keys remain in the workspace. On a persistent box this leaks key material and can break subsequent jobs. Composite actions have no `post:` — add an `if: always()` cleanup step in `deploy.yml`, and append to the search list instead of replacing it.
- **Installer profiles now always active:** removing the `JENKINS_HOME` activation means every local `mvn` run on macOS/Windows also builds and signs the installers. `cli/osx` is covered by the new `local` default profile, but the `osx` module now picks the Developer ID certificate without `-Pdebug`, which developers without the cert don't have. Possibly intended, but a noticeable behavior change for local builds.
- Minor: `beta.yml` is still `name: Publish Snapshot` (collides with `snapshot.yml` in the Actions UI); the ED25519 step inlines the secret via `${{ env.… }}` instead of `printf '%s' "$VAR"` like the DSA step (fragile with multi-line PEMs); `appstore.yml` triggers on *every* tag push, same as `release.yml` (intended?); `xcrun altool --upload-app` is deprecated — likely still works with the pinned Xcode 15 (`DEVELOPER_DIR`, which is missing from the appstore job), but Transporter/App Store Connect API is the supported path going forward.

**Suggested order:** fix the `runs-on` expression and `!cancelled()` first, then run a manual `beta.yml` dispatch — that exercises Windows signing (token/PIN), the `archive: false` issues, and the feed uploads in one pass without risking a real release.

.github/workflows/appstore.yml
```diff
@@ -0,0 +1,34 @@
...
+      SPARKLE_FEED: ''
+      DEVELOPER_DIR: '/Applications/Xcode 26.app/Contents/Developer'
+    secrets: inherit
+
+  appstore:
+    needs: deploy
+    permissions:
+      actions: read
+      contents: read
+    runs-on: macos
```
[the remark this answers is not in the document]
dkocher: Using self-hosted runner labels.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,133 @@
...
+        required: true
+        type: string
+
+jobs:
+  publish:
+    if: ${{ !cancelled() }}
+    permissions:
+      actions: read
+      contents: read
+    runs-on: linux
```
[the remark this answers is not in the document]
dkocher: Using self-hosted runner labels.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,133 @@
...
+
+      - name: Credentials
+        id: identity
+        env:
+          SFTP_IDENTITY: ${{ secrets.SFTP_IDENTITY }}
+        shell: bash
+        run: |
+          printf '%s' "$SFTP_IDENTITY" > private-key.pem
+          echo "KEYFILE=private-key.pem" >> "$GITHUB_OUTPUT"
+
```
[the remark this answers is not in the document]
dkocher: In 03deae77.

.github/actions/import-developer-id-certificate/action.yml
```diff
@@ -0,0 +1,35 @@
...
+        printf '%s' "$CERTIFICATE_BASE64" | base64 -D > certificate.p12
+        KEYCHAIN_PATH=${{ runner.temp }}/build.keychain-db
+        KEYCHAIN_PASSWORD=$(uuidgen)
+        /usr/bin/security create-keychain -p "$KEYCHAIN_PASSWORD" "$KEYCHAIN_PATH"
+        /usr/bin/security set-keychain-settings -u "$KEYCHAIN_PATH"
+        /usr/bin/security unlock-keychain -p "$KEYCHAIN_PASSWORD" "$KEYCHAIN_PATH"
+        curl -sS -o AppleWWDRCAG3.cer https://www.apple.com/certificateauthority/AppleWWDRCAG3.cer
+        /usr/bin/security import AppleWWDRCAG3.cer -k "$KEYCHAIN_PATH" -T /usr/bin/codesign
+        curl -sS -o DeveloperIDCA.cer https://www.apple.com/certificateauthority/DeveloperIDCA.cer
+        /usr/bin/security import DeveloperIDCA.cer -k "$KEYCHAIN_PATH" -T /usr/bin/codesign
```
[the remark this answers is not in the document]
dkocher: In dcfaa8f3.

.github/workflows/deploy.yml
```diff
@@ -26,34 +40,142 @@ jobs:
...
+        shell: bash
+        run: |
+          printf '%s' "$SPARKLE_PRIVATE_KEY_DSA" > www/update/private.pem
+        env:
+          SPARKLE_PRIVATE_KEY_DSA: ${{ secrets.SPARKLE_PRIVATE_KEY_DSA }}
+      - name: Copy Sparkle Updater Private Key (ED25519)
+        shell: bash
+        run: echo "${{ env.SPARKLE_PRIVATE_KEY_ED25519 }}" > www/update/private-ed25519.pem
+        env:
+          SPARKLE_PRIVATE_KEY_ED25519: ${{ secrets.SPARKLE_PRIVATE_KEY_ED25519 }}
```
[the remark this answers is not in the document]
dkocher: In e66229f0.

.github/workflows/deploy.yml
```diff
@@ -26,34 +40,142 @@ jobs:
...
         with:
-          args: install bonjour -y
-      - name: Install WiX Toolset
-        if: ${{ runner.os == 'Windows' }}
-        uses: crazy-max/ghaction-chocolatey@v4
+          archive: false
+          path: |-
+            osx/target/release/*.zip
+      - name: Archive App PKG (macOS)
+        uses: actions/upload-artifact@v7
```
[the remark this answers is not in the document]
dkocher: Default is 'warn' according to https://github.com/actions/upload-artifact#inputs

dkocher:
Depends on #18407 to test.

build.xml
```diff
@@ -148,11 +151,17 @@
 
     <target name="msbuild-props" depends="version">
         <condition property="msbuild.sign" value="false">
-            <isset property="skipSign"/>
+            <isset property="env.SKIP_SIGN"/>
```
ylangisc: Should be explicitly check if `true` as with GitHub workflow a boolean input that is not set defaults to `false` and would thus skip signing.

→ `<istrue value="${env.SKIP_SIGN}"/>`

osx/build.xml
```diff
@@ -42,12 +42,16 @@
               value="CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO JVM_RUNTIME_ARGS='${jvm.runtime.debug}${jvm.runtime.args}' JAVA_HOME=${java.home} SYMROOT=${build} SDKROOT=macosx ARCHS=${app.runtime.archs} MACOSX_DEPLOYMENT_TARGET=${app.runtime.system.min}"/>
 
     <condition property="build.xcodeoptions"
-               value="${build.xcodeoptions.default} SPARKLEFEED='https://version.cyberduck.io/${sparkle.feed}/changelog.rss' INFOPLIST_PREPROCESSOR_DEFINITIONS=SPARKLE_BUILD"
+               value="${build.xcodeoptions.default} SPARKLEFEED='https://version.cyberduck.io/${env.SPARKLE_FEED}/changelog.rss' INFOPLIST_PREPROCESSOR_DEFINITIONS=SPARKLE_BUILD"
                else="${build.xcodeoptions.default}">
-        <isset property="sparkle.feed"/>
+        <isset property="env.SPARKLE_FEED"/>
```
ylangisc: `<not><equals arg1="${env.SPARKLE_FEED}" arg2=""/></not>`. Same reason as for `SKIP_SIGN`, in this case for the App Store build.  Alternatively, tie `SPARKLE_BUILD` explicitly to `XCODE_CONFIGURATION != 'Mac App Store'`.

.github/workflows/deploy.yml
```diff
@@ -37,23 +79,162 @@ jobs:
...
+          path: |-
+            cli/osx/target/release/*.pkg
+      - name: Archive CLI Formula (macOS)
+        if: runner.os == 'macos'
+        uses: actions/upload-artifact@v7
+        with:
+          archive: false
+          path: |-
+            cli/osx/target/release/duck.rb
+      - name: Archive EXE (Windows)
```
ylangisc, line 174: `Archive CLI Tarball (macOS)` is missing for `cli/osx/target/release/*.tar.gz`.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,230 @@
...
+  cli:
+    permissions:
+      actions: read
+      contents: read
+    runs-on: [ linux, x64 ]
+    steps:
+      - name: Download CLI
+        uses: actions/download-artifact@v8
+        with:
+          pattern: '{duck-*.pkg,duck-*.exe,duck-*.msi}'
```
ylangisc: Missing `duck-*.tar.gz`

.github/workflows/snapshot.yml
```diff
@@ -1,14 +1,113 @@
...
+    if: ${{ !cancelled() }}
+    uses: ./.github/workflows/publish.yml
+    with:
+      SPARKLE_FEED: nightly
+    secrets: inherit
+
+  # Update the Homebrew tap at github.com/iterate-ch/homebrew-cyberduck. Nightly
+  # only, so it lives here rather than in the shared publish.yml.
+  brew:
+    needs: deploy
```
ylangisc: Runs parallel to publish. Need to wait for publish too.

.github/workflows/snapshot.yml
```diff
@@ -1,14 +1,113 @@
...
+          USERNAME: '${{ secrets.SFTP_USER }}'
+          IDENTITY: '${{ steps.identity.outputs.KEYFILE }}'
+        with:
+          mode: upload
+          url: sftp://update.cyberduck.io/var/data/update.cyberduck.io/nightly/
+          path: release/Cyberduck-*
+          args: '--existing overwrite --verbose'
+
+  publish:
+    needs: [ deploy, sftp ]
```
ylangisc, line 73: Only waits for the needed jobs to finish, regardless of their result. A failed sftp job still publishes the appcast.

osx/pom.xml
```diff
@@ -229,40 +228,10 @@
...
-            <properties>
-                <xcode.configuration>Mac App Store</xcode.configuration>
-                <codesign.provisionprofile.name>appstore.provisionprofile</codesign.provisionprofile.name>
-                <skipNotarize>true</skipNotarize>
-            </properties>
         </profile>
         <profile>
             <id>installer</id>
             <activation>
-                <property>
```
ylangisc, line 263: Removing this activation will make the package target run for local builds too. Bind activation to `<property><name>env.CI</name></property>`. `CI` is always set for GitHub actions.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,230 @@
...
+          merge-multiple: true
+
+      - name: Rackspace Upload
+        uses: iterate-ch/cyberduck-cli-action@v1
+        env:
+          USERNAME: ${{ secrets.RACKSPACE_USER }}
+          PASSWORD: ${{ secrets.RACKSPACE_PASSWORD }}
+        with:
+          mode: upload
+          url: ${{ inputs['SPARKLE_FEED'] == 'nightly' && 'rackspace:/cdn.duck.sh/nightly/' || inputs['SPARKLE_FEED'] == 'beta' && 'rackspace:/cdn.duck.sh/nightly/' || 'rackspace:/cdn.duck.sh/' }}
```
ylangisc: For beta i'ts uploaded to nightly URL.

cli/windows/build.xml
```diff
@@ -68,7 +68,7 @@
             <include name="tools/*.ps1"/>
             <replacefilter token="${VERSION}" value="${version}"/>
             <replacefilter token="${REVISION}" value="${revision}"/>
-            <replacefilter token="${FEED}" value="${sparkle.feed}"/>
+            <replacefilter token="${FEED}" value="${env.SPARKLE_FEED}"/>
```
ylangisc: `${FEED}` now resolves to `release` here, so the Chocolatey `duck` package downloads from `https://dist.duck.sh/release/duck-….exe` (`setup/chocolatey/duck/tools/chocolateyInstall.ps1`). Release CLI installers are uploaded to the container root though: `dist.duck.sh/duck-9.5.3.45464.exe` works, `dist.duck.sh/release/duck-9.5.3.45464.exe` is a 404. `choco install duck` would fail for releases. Omit the feed segment for `release`?

.github/workflows/release.yml
```diff
@@ -1,14 +1,158 @@
...
+        uses: actions/download-artifact@v8
+        with:
+          pattern: Cyberduck-*
+          path: release
+          merge-multiple: true
+
+      - name: Download CLI
+        uses: actions/download-artifact@v8
+        with:
+          pattern: duck-*
```
ylangisc: `jpackage` names Debian packages `duck_` thus the pattern should be`'{duck-*,duck_*}'`.

.github/workflows/release.yml
```diff
@@ -1,14 +1,158 @@
...
+          files: |
+            release/Cyberduck-*.zip
+            release/Cyberduck-*.pkg
+            release/Cyberduck-*.exe
+            release/Cyberduck-*.msi
+            release/Cyberduck-*.appx
+            release/duck-*.pkg
+            release/duck-*.exe
+            release/duck-*.msi
+            release/duck-*.deb
```
ylangisc: `jpackage` names Debian packages `duck_*.deb`.

.github/workflows/build.yml
```diff
@@ -28,7 +28,7 @@ jobs:
         with:
           fetch-depth: 0
       - name: Set up JDK 21
-        uses: actions/setup-java@v6
+        uses: actions/setup-java@v5
```
ylangisc, line 31: What is the reason for this rever (b4b7e432f0)? Consider a dependabot ignore config if there is valid reason.

osx/build.xml
```diff
@@ -157,8 +161,8 @@
             <include name="changelog.rss"/>
             <replacefilter token="${TIMESTAMP}" value="${touch.time}"/>
             <replacefilter token="${ARCHIVE}"
-                           value="https://update.cyberduck.io/${sparkle.feed}/${bundle.archive.zip}"/>
-            <replacefilter token="${CHANGELOG}" value="https://version.cyberduck.io/${sparkle.feed}/changelog.html"/>
+                           value="https://update.cyberduck.io/${env.SPARKLE_FEED}/${bundle.archive.zip}"/>
+            <replacefilter token="${CHANGELOG}" value="https://version.cyberduck.io/${env.SPARKLE_FEED}/changelog.html"/>
```
ylangisc, line 167: Jenkins used an empty feed name for releases (current appcast: `https://version.cyberduck.io//changelog.html`), while `release.yml` now passes `SPARKLE_FEED=release`. Feed and enclosure paths are covered by the server-side aliases (`/macos/release/`, `/windows/release/`, `update.cyberduck.io/release/`), but this release notes link becomes `https://version.cyberduck.io/release/changelog.html`, which is a 404. Suggest `https://version.cyberduck.io/macos/${env.SPARKLE_FEED}/changelog.html`, consistent with Windows. Same root cause as the Chocolatey `dist.duck.sh/release/` path (https://github.com/iterate-ch/cyberduck/pull/17126#discussion_r3988463924).

ylangisc:
Mermaid diagrams of the workflow setup in a README.md would be beneficial.

.github/workflows/build.yml
dkocher: Checksum validation failures now enabled as described in https://github.com/actions/setup-java#download-integrity-and-signatures failed.

.github/workflows/deploy.yml
dkocher: In c68e3d74.

.github/workflows/publish.yml
dkocher: In 5a79345d.

.github/workflows/release.yml
dkocher: In 5a79345d.

.github/workflows/release.yml
dkocher: In 5a79345d.

.github/workflows/publish.yml
dkocher: This is intended.

osx/pom.xml
dkocher: In ca46d07f.

build.xml
dkocher: In 97dfc635.

dkocher:
> Mermaid diagrams of the workflow setup in a README.md would be beneficial.

Not sure if that adds any real value

```mermaid
flowchart TD
    push_master(["push → master"])
    push_tag(["push tag"])
    pull_request(["pull_request"])
    dispatch(["workflow_dispatch"])
    cron(["nightly cron"])

    pull_request --> buildYml
    push_master --> buildYml
    buildYml[["build.yml<br/>Unit Tests"]]

    cron --> ciYml
    dispatch --> ciYml
    ciYml[["ci.yml<br/>Integration Tests"]]

    dispatch --> transifexYml
    transifexYml[["transifex.yml<br/>Pull Localizations"]]

    push_master --> snapshotYml
    snapshotYml[["snapshot.yml<br/>Publish Snapshot"]]

    dispatch --> betaYml
    betaYml[["beta.yml<br/>Publish Beta"]]

    push_tag --> releaseYml
    releaseYml[["release.yml<br/>Publish Release"]]

    push_tag --> appstoreYml
    appstoreYml[["appstore.yml<br/>Publish Mac App Store"]]

    push_tag --> imageYml
    imageYml[["image.yml<br/>Docker Image (CLI)"]]

    subgraph Reusable["Reusable workflows"]
        deployYml[["deploy.yml<br/>matrix: macOS/arm64, Windows/x64, Linux/x64+arm64"]]
        publishYml[["publish.yml<br/>app · cli · changelog · deb · rpm"]]
    end

    snapshotYml -->|SPARKLE_FEED=nightly| deployYml
    betaYml -->|SPARKLE_FEED=beta| deployYml
    releaseYml -->|SPARKLE_FEED=release| deployYml
    appstoreYml -->|profile=appstore| deployYml

    snapshotYml -->|SPARKLE_FEED=nightly| publishYml
    betaYml -->|SPARKLE_FEED=beta| publishYml
    releaseYml -->|SPARKLE_FEED=release| publishYml

    snapshotYml --> snapshotSftp["sftp job<br/>update.cyberduck.io/nightly"]
    snapshotYml --> brewJob["brew job<br/>homebrew-cyberduck tap"]
    snapshotSftp -.must finish before.-> publishYml

    releaseYml --> releaseSftp["sftp job<br/>dist.duck.sh"]
    releaseYml --> ghRelease["release job<br/>GitHub Release + assets"]
    releaseYml --> chocoJob["chocolatey job<br/>Chocolatey push"]
```

osx/build.xml
dkocher: In 7d74cae5.

.github/workflows/snapshot.yml
dkocher: In f3d00061.

cli/windows/build.xml
dkocher: In 258f5b00.

osx/build.xml
dkocher: In 258f5b00.

.github/workflows/snapshot.yml
dkocher: In 72a3aa96.

dkocher: This is just an additional upload. Build will be pushed in publish job to CDN:

.github/workflows/publish.yml
dkocher: In dce3ca12.

.github/workflows/build.yml
dkocher: Reapplied in 2dca92322f3d41785dc24110ef3c81888fe3c2d6.

ylangisc (approved):

PR #17132, opened 2025-05-26
Fix goto file

chizkiyahu:
fix https://github.com/iterate-ch/cyberduck/issues/17131

what is do 
1. trim the path
2. in case the path is file open the folder and select the file

### DIFF
--- a/osx/src/main/java/ch/cyberduck/ui/cocoa/controller/GotoController.java
+++ b/osx/src/main/java/ch/cyberduck/ui/cocoa/controller/GotoController.java
@@ -39,6 +39,7 @@ import org.rococoa.cocoa.foundation.NSInteger;
 import org.rococoa.cocoa.foundation.NSRect;
 
 import java.util.Comparator;
+import java.util.EnumSet;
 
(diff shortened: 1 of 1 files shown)


dkocher:
We do not accept pull requests from [_ChatGPT Connector_](https://github.com/apps/chatgpt-connector). Any fix would require
* Fix both macOS and Windows implementations
* Not depend on `java.nio` API unrelated for remote file paths

Chizkiyahu:
@dkocher 
I am a python devloper that use cyberduck every day 
this code created by me (with help of ai) but every line in this code are writtern by me 
its take me a few houres to make this pr 
I do it to solved the bug for my self (and its working for me )
and open the PR to contrubute back to the community 

about the windows fails - I check the error and found `Address already in use: bind` 
this look like to me as a problem that rerun can fix and is not connected to the code that i changed 
I also run the [same tests on my repo and its pass](https://github.com/Chizkiyahu/cyberduck/actions/runs/15280745183/job/42978863960)

about `java.io` 
what API i needs to use to check if path is a file or folder 

thanks

PR #17145, opened 2025-06-02
Update dependency and load native libraries without unpacking from archive

### LINKED ISSUE #17143: Update jansi dependency
- [ ] Update to version with arm64 support
- [ ] Unpack native libraries from official distribution
- [ ] Add dependencies for native libraries
- [ ] Set `library.jansi.path` in deployment

dkocher:
Fix #17143.

### DIFF
--- a/cli/dll/pom.xml
+++ b/cli/dll/pom.xml
@@ -130,5 +130,21 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
+            <groupId>org.fusesource.jansi</groupId>
+            <artifactId>jansi</artifactId>
+            <version>${jansi.version}</version>
+            <classifier>x86_64</classifier>
+            <type>dll</type>
+            <scope>runtime</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.fusesource.jansi</groupId>
+            <artifactId>jansi</artifactId>
+            <version>${jansi.version}</version>
+            <classifier>arm64</classifier>
+            <type>dll</type>
+            <scope>runtime</scope>
+        </dependency>
     </dependencies>
 </project>
--- a/cli/linux/pom.xml
+++ b/cli/linux/pom.xml
@@ -81,6 +81,14 @@
                     <version>${jna-version}</version>
                     <scope>runtime</scope>
                 </dependency>
+                <dependency>
+                    <groupId>org.fusesource.jansi</groupId>
+                    <artifactId>jansi</artifactId>
+                    <version>${jansi.version}</version>
+                    <classifier>arm64</classifier>
+                    <type>so</type>
+                    <scope>runtime</scope>
+                </dependency>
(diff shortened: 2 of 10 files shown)


cli/linux/pom.xml
```diff
@@ -127,5 +127,21 @@
...
+            <scope>runtime</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.fusesource.jansi</groupId>
+            <artifactId>jansi</artifactId>
+            <version>${jansi.version}</version>
+            <classifier>arm64</classifier>
+            <type>so</type>
+            <scope>runtime</scope>
+        </dependency>
```
AliveDevil: - Move dependencies to profile matching architecture.
- armhf runtime missing.
- Missing initialization for CLI

cli/dll/pom.xml
```diff
@@ -182,5 +182,21 @@
...
+            <scope>runtime</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.fusesource.jansi</groupId>
+            <artifactId>jansi</artifactId>
+            <version>${jansi.version}</version>
+            <classifier>arm64</classifier>
+            <type>dll</type>
+            <scope>runtime</scope>
+        </dependency>
```
AliveDevil, line 148: arm64, x86_64 copy to the same path - thus one overwrites the other.

AliveDevil (changes requested):

cli/dll/pom.xml
AliveDevil: As duck CLI v9 doesn't support arm64, remove the dependency here.

cli/linux/pom.xml
dkocher: In 1e3087c9

dkocher: In 66671521

dkocher:
Build failure on Windows

```
[INFO]   [msbuild]   D:\a\cyberduck\cyberduck\core\dll\Cyberduck.Core.ikvmproj : error MSB4236: The SDK 'IKVM.NET.Sdk' specified could not be found.
```

cli/osx/pom.xml
```diff
@@ -226,5 +226,21 @@
...
+            <scope>runtime</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.fusesource.jansi</groupId>
+            <artifactId>jansi</artifactId>
+            <version>${jansi.version}</version>
+            <classifier>arm64</classifier>
+            <type>jnilib</type>
+            <scope>runtime</scope>
+        </dependency>
```
dkocher: Will need to add as universal binary instead.

dkocher: Resolved with 340a3ced6f86964244fbcdb4b1e7af3cfc6fa7ce

AliveDevil (approved):

PR #17154, opened 2025-06-04
Refactor alerts to allow display as sheet or modal.

dkocher:
Fix #17147.


dkocher:
Drops the custom panel for the login prompt with a standard alert.
Screenshot 2025-06-13 at 17 58 19
Screenshot 2025-06-13 at 17 59 44

PR #17226, opened 2025-07-03
Access to the directory id cache must be synchronized.

ylangisc:


dkocher (commented):
Could this be simplified using a loading cache `LRUCache.usingLoader`?

ylangisc:
As we pre-set the `directoryId` in the bulk feature the interface is not very intuitive and not suitable for a loader. Interface is already changed on the UVF branch where a loader can be used then.

dkocher (approved):

PR #17255, opened 2025-07-18
Always scroll to recently added transfer

### LINKED ISSUE #17251: Transfer window does not scroll to show latest transfer

 * If you have trouble connecting to a server or your login credentials are not valid, try to resolve the issue with the assistance of your hosting service provider. Any such ticket will be closed with a resolution of _thirdparty_.

**Describe the bug**
Using the GUI, when I first start transferring in a session, the transfer completion message is visible in the transfer window. After several transfers, the transfer window is filled, and subsequent transfers are no longer visible. Having to scroll after each transfer to confirm the status is annoying.

**To Reproduce**
1. Use the GUI to initiate transfers.
2. Set the transfer window to a short height.
3. Perform several transfers until the transfer window needs to scroll.
4. Note that the most recent transfer is not visible in the window without manual scrolling.

**Expected behavior**
I expect the most recent transfer to always be visible in the log window. For example:
- Order transfers by most-recent-first, and always insert at top, _or_
- Keep the transfers in oldest-first order, but always scroll to the bottom when a new transfer is added.

**Screenshots**
n/a

 - OS: Windows on ARM version: 10.0.26200.5702
 - Version 9.1.7 (43306)

**Log Files**
n/a

**Additional context**

AliveDevil:
See #17251 
User intent is to start a transfer, this transfer should be brought into visible viewport.
Selection will only be performed when nothing has been selected prior.

### DIFF
--- a/windows/src/main/csharp/ch/cyberduck/ui/ViewModels/TransfersViewModel.cs
+++ b/windows/src/main/csharp/ch/cyberduck/ui/ViewModels/TransfersViewModel.cs
@@ -424,11 +424,7 @@ public sealed partial class TransfersViewModel : ObservableObject, IDisposable
 
     private void OnRevealTransfer(TransferViewModel transfer)
     {
-        if (SelectedTransfers.Count is 0)
-        {
-            SelectedTransfer = transfer;
-        }
-        
+        SelectedTransfer = transfer;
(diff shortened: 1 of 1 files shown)


dkocher (changes requested):
On macOS we select and scroll to the added transfer regardless of the current selection.

dkocher (approved):

PR #17311, opened 2025-08-07
 Set QuickXorHash as checksum to compare file changes

dkocher:
~~Extract UUID from ETag formatted as `"{D7751A14-FEA5-4BDA-969E-F8022E58B7E7},9"` which remains constant when moving files.~~


AliveDevil:
We have Hashes for Files available already:
https://github.com/iterate-ch/onedrive-java-client/blob/develop/src/main/java/org/nuxeo/onedrive/client/types/File.java#L9-L11

quickXorHash is guaranteed to be available.

AliveDevil:
Alternatively, use the cTag[^1]-property:
>An eTag for the content of the item. This eTag isn't changed if only the metadata is changed. Note This property isn't returned if the item is a folder. Read-only.

[^1]: https://learn.microsoft.com/en-us/graph/api/resources/driveitem?view=graph-rest-1.0#properties

Relying on a implementation detail of the eTag is dangerous.

core/src/main/java/ch/cyberduck/core/io/Checksum.java
```diff
@@ -65,41 +73,58 @@ public Checksum(final Checksum other) {
...
-                break;
-            case 128:
-                if(hash.matches("[A-Fa-f0-9]{128}")) {
-                    return new Checksum(HashAlgorithm.sha512, hash);
-                }
-                break;
-            default:
-                log.warn("Failure to detect algorithm for checksum {}", hash);
+        // Check for Base64 with propper padding
+        if(hash.matches("^[A-Za-z0-9+/]+=*$") && hash.length() % 4 == 0) {
```
ylangisc: Only 2 `=` are allowed → `^[A-Za-z0-9+/]+={0,2}$`

onedrive/src/main/java/ch/cyberduck/core/onedrive/features/GraphAttributesFinderFeature.java
```diff
@@ -89,6 +92,13 @@ private DriveItem.Metadata toMetadata(final Path file, final DriveItem item) thr
     public PathAttributes toAttributes(final DriveItem.Metadata metadata) {
         final PathAttributes attributes = new PathAttributes();
         attributes.setETag(metadata.getETag());
+        final File file = metadata.getFile();
+        if(file != null) {
+            final Hashes hashes = file.getHashes();
+            if(hashes != null) {
+                attributes.setChecksum(Checksum.parse(hashes.getQuickXorHash()));
```
AliveDevil, line 99: This relies on quickXorHash always being 20 Bytes, and is misrepresented as SHA1.

dkocher: You are right. Full implementation requires us to implement `quickXorHash` as a `ChecksumCompute` feature.

Issue #17318, opened 2025-08-11
Skip listing OneDrive Personal Vault


AliveDevil:
Microsoft Graph doesn't expose the "containingDrivePolicyScenarioViewpoint" field on "remoteItem" metadata.
Use "api.onedrive.com", which does expose this field.

AliveDevil:
Apparently[^1] this is exposed when using
```
Prefer: Include-Feature=Vault
```
When not specifically excluding this feature (i.e. it is active by-default for missing `Prefer` or empty `Prefer: Include-Feature=`) it will show the vault.
Set HTTP Header `Prefer: Include-Feature=AddToOneDrive` to skip listing the vault.

[^1]: [OneDrive/onedrive-api-docs#1427](https://redirect.github.com/OneDrive/onedrive-api-docs/issues/1427#issuecomment-1489045908)

PR #17319, opened 2025-08-11
Filter bogus Item in OneDrive listings

AliveDevil:
SharedWithMeListService can return empty ParentReference for an item
Filtering in base implementation for any empty file id.


dkocher (approved):

dkocher:
Relates to #17318?

ylangisc (approved):

AliveDevil:
Doesn’t relate to that, no.
17318 is about being able to filter out the Personal Vault.
This one is about Graph returning items without Ids.

PR #17322, opened 2025-08-11
Set ctag as version id for files for comparison.

dkocher:


dkocher:
Depends on #17312.

onedrive/src/main/java/ch/cyberduck/core/onedrive/GraphSession.java
```diff
@@ -96,6 +96,7 @@ public ODataQuery getQuery(ODataQuery query) {
 
                 // Drive Item properties
                 DriveItem.Property.File, /*Usage: Determines File */
+                DriveItem.Property.CTag,
```
AliveDevil:
```suggestion
                DriveItem.Property.CTag, /*Usage: File Checksum/Version Comparison */
```

AliveDevil:
No dependency on #17312, 3.2.4 was a typo fix for quickXorHash in Hashes-facet.

ylangisc (approved):

PR #17325, opened 2025-08-12
Use QuickXorHash for version identifier as ctag apparently changes on…

dkocher:
… rename.


AliveDevil (approved):
Microsoft (╯‵□′)╯︵┻━┻

PR #17327, opened 2025-08-12
Stop list of PrivateKeys to append keys to not show duplicates

### LINKED ISSUE #17320: List of SSH Private Keys gets filled with repeats

 * If you have trouble connecting to a server or your login credentials are not valid, try to resolve the issue with the assistance of your hosting service provider. Any such ticket will be closed with a resolution of _thirdparty_.

**Describe the bug**
When I want to connect to my Ubuntu System via Cyberduck I get asked for the username, password and a possible SSH Private Key

Image

However if login fails and I try to login again the list of private keys gets appended instead of cleared so it shows repeated keys. This gets repeated every time a login fails so the list grows longer and longer

Image

Image

**To Reproduce**
1. Try to connect to a SFTP session
2. When asked for login and password enter wrong information so the login fails
3. Open the list of Private SSH Keys

**Expected behavior**
The list should only show the possible keys once and not multiple times

**Screenshots**
Shown above
 - OS: Windows 10
 - Version22H2

**Log Files**
Please include 

[cyberduck.log](https://github.com/user-attachments/files/21715943/cyberduck.log)

the [log file](https://docs.cyberduck.io/cyberduck/support/#logging-output). You can reach the logging output in _Preferences → Connection_.

**Additional context**

Akashic101:


windows/src/main/csharp/ch/cyberduck/ui/controller/PromptLoginController.cs
```diff
@@ -118,6 +118,9 @@ public Local select(Local identity)
         }
 
         private void InitPrivateKeys()
+
+        _keys.Clear();
+
```
AliveDevil: Wrong location for method body.

AliveDevil (changes requested):

Akashic101 (commented):
Moved clearing of the keys actually into the function

AliveDevil (approved):

PR #17400, opened 2025-08-28
Fix mapping of tags to option

### LINKED ISSUE #17358: Upload "Cancel" button still performs the upload
**Describe the bug**
When I drag and drop a file to an SFTP folder I see an Overwrite window with the options of "Cancel" or "Continue".  If I click "Cancel" the file will upload and overwrite the file.

**To Reproduce**
For this example, I have an existing file in the SFTP location with the same name as the upload.
1. Open an SFTP folder in Cyberduck
2. Drag and drop the local file to the Cyberduck window
3. When shown the Overwrite window, click "Cancel"
4. File upload continues as normal and overwrites the existing file.

**Expected behavior**
I would expect "Cancel" in the Overwrite window, it would stop the upload process.

**Screenshots**

 - OS: Mac OS
 - Version: 15.5

**Log Files**
Unable to find recent log file. The one I found (from console.app) only had entries from May 2024.

**Additional context**
I am using an AWS S3 connection to access my SFTP folders.

### LINKED ISSUE #17366: Cancel New connection window doesn't work at all
**Describe the bug**
Open connection doesn't close

**To Reproduce**
1. Go to 'Open connection'
2. Click on 'Cancel'
3. See error

**Expected behavior**
Connection window should close

**Screenshots**
Image

 - OS: [macOS]
 - Version [MB Pro M$ Sequoia 15.6]

dkocher:


OlliStox:
In https://github.com/iterate-ch/cyberduck/issues/17358 the bug is marked as closed, but it is not solved in 9.2.1.

PR #17491, opened 2025-09-22
Sanitize User-Agent HTTP header

AliveDevil:
[RFC9110](https://www.rfc-editor.org/rfc/rfc9110#section-5.6.2) requires the [User-Agent](https://www.rfc-editor.org/rfc/rfc9110#section-10.1.5) header to be well-formatted.
Currently, "Mountain Duck" violates this principle, as whitespace is not a valid token.

### DIFF
--- a/core/src/main/java/ch/cyberduck/core/PreferencesUseragentProvider.java
+++ b/core/src/main/java/ch/cyberduck/core/PreferencesUseragentProvider.java
@@ -40,20 +40,20 @@ public class PreferencesUseragentProvider implements UseragentProvider {
     }
 
     private static String sanitizeApplicationName(final String application) {
-        final char[] rfc2616TokenSeparators = new char[]{
-                '(', ')', '<', '>', '@',
-                ',', ';', ':', '\\', '"',
-                '/', '[', ']', '?', '=',
-                '{', '}', ' ', '\t'
+        final char[] httpHeaderTokenSeparators = new char[]{
+                ' ', '\t', '"', '(', ')',
+                ',', '/', ':', ';', '<',
+                '=', '>', '?', '@', '[',
+                '\\', ']', '{', '}'
         };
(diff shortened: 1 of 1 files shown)


ylangisc (approved):

core/src/main/java/ch/cyberduck/core/PreferencesUseragentProvider.java
```diff
@@ -21,18 +21,47 @@
...
+        }
 
-    private static final String ua = String.format("%s/%s.%s (%s/%s) (%s)",
-            preferences.getProperty("application.name"),
-            preferences.getProperty("application.version"),
-            preferences.getProperty("application.revision"),
-            preferences.getProperty("os.name"),
-            preferences.getProperty("os.version"),
-            preferences.getProperty("os.arch"));
+        final StringBuilder builder = new StringBuilder(application.length() - 1);
```
dkocher: Optimization is probably futile, Can initialize with default constructor.

core/src/main/java/ch/cyberduck/core/PreferencesUseragentProvider.java
```diff
@@ -21,18 +21,47 @@
...
+        for(int i = 0; i < application.length(); i++) {
+            final char token = application.charAt(i);
+            if(ArrayUtils.indexOf(httpHeaderTokenSeparators, token) != -1) {
+                continue;
+            }
+
```
dkocher: Remove line break.

core/src/main/java/ch/cyberduck/core/PreferencesUseragentProvider.java
```diff
@@ -21,18 +21,47 @@
...
+            builder.append(token);
+        }
+
```
dkocher: Remove line break.

dkocher (changes requested):
Nitpicking.

dkocher (approved):

PR #17515, opened 2025-09-26
Feature/log4j2 plugin

ylangisc:


core/dylib/src/main/java/ch/cyberduck/core/logging/SystemLogAppender.java
```diff
@@ -37,10 +40,14 @@
 /**
  * Redirect to NSLog(). Logs an error message to the Apple System Log facility.
  */
+@Plugin(
+        name = "UnifiedSystemLog",
```
dkocher: Should be different name?

dkocher (changes requested):

dkocher:
No [annotation processing](https://logging.apache.org/log4j/2.x/manual/plugins.html#plugin-registry) required?

ylangisc:
> No [annotation processing](https://logging.apache.org/log4j/2.x/manual/plugins.html#plugin-registry) required?

No, annotation has `Runtime` retention policy.

dkocher (approved):

Issue #17612, opened 2025-11-03
Support HTTP Credential URL for AWS/S3 in Cyberduck


sannies:
**Is your feature request related to a problem? Please describe.**
In many AWS workflows today, credentials are not static. Instead, a temporary credential issuance service runs behind a custom HTTP endpoint. The official AWS SDKs (e.g., JavaScript v3’s fromHttp() provider) support this pattern.
Cyberduck currently lacks the ability to fetch AWS credentials via an arbitrary HTTP endpoint and refresh them automatically.
Existing issues such as #8610, #10989, #14862 revolve around limitations in metadata or profile-based credential retrieval—but none address a generic HTTP credential endpoint.

**Describe the solution you'd like**
I implemented a solution that adds a new bookmark property in the Custom dict:

```

    <key>Custom</key>
    <dict>
        <key>aws.credentials.http.url</key>
       <string>https://up.content.example.com/api_v1/upload/VMHGTtd0BDYpcZgWU2arm6SwElXsiOCK</string>
    </dict>
```
When present:

- `S3CredentialsConfigurator` detects the property and returns placeholder credentials so that validation doesn’t block connection.
- `S3Session.configureCredentialsStrategy` skips other auth methods and creates an AWSSessionCredentialsRetriever with this URL.
- `AWSSessionCredentialsRetriever` has been modified to:
   - Use the provided URL literally (preserving path and encoding)
   - Perform HTTP GET and parse AWS STS-style JSON:
     
```
{ "AccessKeyId": "...", "SecretAccessKey": "...", "Token": "...", "Expiration": "..." }
```
If the property is not set, Cyberduck behaves exactly as before (profiles, metadata, etc.).

**Describe alternatives you've considered**
- Using EC2 Instance Metadata (169.254.169.254) as a proxy
It is technically possible to run a local service (or SSH tunnel) that mimics the AWS metadata endpoint and returns credentials. However, this requires routing tricks, hosts file changes, root privileges, or VPN manipulation — fragile and environment-specific.
- Manually pasting temporary AWS credentials into Cyberduck
This works only for short periods until credentials expire. It is error-prone, breaks automation, and is not feasible for non-technical users or expiring STS credentials in CI / upload pipelines.
- Writing credentials into ~/.aws/credentials using AWS CLI or scripts
This requires storing STS tokens in plaintext on disk and still risks expiration. It also makes Cyberduck pick them up only if profile names and hostnames are aligned, which is not always the case.

**Additional context**
[...]

PR #17613, opened 2025-11-03
Http credential provider

sannies:
fixes issue #17612

### DIFF
--- a/defaults/src/main/resources/default/log4j.xml
+++ b/defaults/src/main/resources/default/log4j.xml
@@ -15,13 +15,26 @@
   -->
 <Configuration>
     <Appenders>
-        <Null name="disabled"/>
+        <File name="file" fileName="/tmp/cyberduck.log" append="false">
+            <PatternLayout pattern="%d [%t] %-5p %c - %m%n"/>
+        </File>
+        <Console name="console" target="SYSTEM_OUT">
+            <PatternLayout pattern="%d [%t] %-5p %c - %m%n"/>
+        </Console>
     </Appenders>
     <Loggers>
-        <Logger name="org.apache.http.wire" additivity="false" level="error"/>
-        <Logger name="org.apache.http.headers" additivity="false" level="error"/>
-        <Root level="error">
-            <AppenderRef ref="disabled"/>
+        <Logger name="org.apache.http.wire" additivity="false" level="debug">
+            <AppenderRef ref="file"/>
+            <AppenderRef ref="console"/>
+        </Logger>
+        <Logger name="org.apache.http.headers" additivity="false" level="debug">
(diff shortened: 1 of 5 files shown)


dkocher:
Considered AI slop.

sannies:
Let me respond to that. 
Yes, this was generated with AI - the code base is quite extensive and it is very, very hard to understand. 

Still: The change in [AWSSessionCredentialsRetriever.java](https://github.com/iterate-ch/cyberduck/pull/17613/files#diff-bd32d8a54b4a1a5657f91039d0ab51a92c86906eb6b9fd3b5c34b3bc30872488) is fixing a problematic behavior.

Unfortunately I cannot supply an actual test as I cannot mock the HttpClient in the `AWSSessionCredentialsRetriever` but a few lines of code that mimic what happens inside the `AWSSessionCredentialsRetriever` show the problematic behavior

```
public void testParseWithEscape() throws Exception {
    final TestProtocol defaultProtocol = new TestProtocol(Scheme.http);
    ProtocolFactory.get().register(defaultProtocol);
    final Host address = new HostParser(ProtocolFactory.get()).get("http://customsource/a%2Fb/secrettoken");
    String actualUrl = new HostUrlProvider().withUsername(false).withPath(true).get(address);
    assertEquals("http://customsource/a%2Fb/secrettoken", actualUrl);
}
```

so in the end the request will be issued against  `http://customsource/a/b/secrettoken` which is obviously not what was intended.

sannies:
@dkocher Would you perhaps describe in a few words how you see that custom HTTP credentials sources could be implemented? I'm happy to give it another shot.

dkocher:
> Still: The change in [AWSSessionCredentialsRetriever.java](https://github.com/iterate-ch/cyberduck/pull/17613/files#diff-bd32d8a54b4a1a5657f91039d0ab51a92c86906eb6b9fd3b5c34b3bc30872488) is fixing a problematic behavior.

I do not understand what problematic behaviour you are referring to. If you see an issue with URI parsing you can show this is in a test for `HostParser` [^1].

[^1]: https://github.com/iterate-ch/cyberduck/blob/master/core/src/test/java/ch/cyberduck/core/HostParserTest.java

dkocher:
> @dkocher Would you perhaps describe in a few words how you see that custom HTTP credentials sources could be implemented? I'm happy to give it another shot.

It should suffice to create your own [connection profile](https://docs.cyberduck.io/protocols/profiles/) like [^1] that has the URL to fetch session credentials using `AWSSessionCredentialsRetriever` set for the `Context` key.

[^1]: https://github.com/iterate-ch/profiles/blob/main/S3%20(Credentials%20from%20Instance%20Metadata).cyberduckprofile

PR #17617, opened 2025-11-04
Expose Context key in profile as property and allow override in bookm…

### LINKED ISSUE #17612: Support HTTP Credential URL for AWS/S3 in Cyberduck

dkocher:
…ark.

### DIFF
--- a/core/src/main/java/ch/cyberduck/core/Profile.java
+++ b/core/src/main/java/ch/cyberduck/core/Profile.java
@@ -685,6 +685,9 @@ public class Profile implements Protocol {
                 property -> StringUtils.contains(property, '=') ? substitutor.replace(StringUtils.substringAfter(property, '=')) : StringUtils.EMPTY)));
         // In profile as dict
         properties.putAll(this.map(PROPERTIES_KEY));
+        if(null != this.getContext()) {
+            properties.put(CONTEXT_KEY, this.getContext());
+        }
         return properties;
     }
 
--- a/s3/src/main/java/ch/cyberduck/core/auth/AWSSessionCredentialsRetriever.java
+++ b/s3/src/main/java/ch/cyberduck/core/auth/AWSSessionCredentialsRetriever.java
@@ -75,7 +75,7 @@ public class AWSSessionCredentialsRetriever implements S3CredentialsStrategy {
         final HttpClientBuilder configuration = builder.build(ProxyFactory.get(),
                 new DisabledTranscriptListener(), new DisabledLoginCallback());
         try (CloseableHttpClient client = configuration.build()) {
-            final HttpRequestBase resource = new HttpGet(new HostUrlProvider().withUsername(false).withPath(true).get(address));
+            final HttpRequestBase resource = new HttpGet(url);
             return client.execute(resource, response -> {
                 switch(response.getStatusLine().getStatusCode()) {
                     case HttpStatus.SC_OK:
--- a/s3/src/main/java/ch/cyberduck/core/s3/S3Session.java
+++ b/s3/src/main/java/ch/cyberduck/core/s3/S3Session.java
@@ -27,6 +27,7 @@ import ch.cyberduck.core.ListService;
 import ch.cyberduck.core.LoginCallback;
 import ch.cyberduck.core.Path;
 import ch.cyberduck.core.PathContainerService;
+import ch.cyberduck.core.Profile;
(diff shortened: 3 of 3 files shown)


sannies:
[look_at_preferences_all_the_time_parse_context_property_prevent_login_dialog_use_configure.patch](https://github.com/user-attachments/files/23357400/look_at_preferences_all_the_time_parse_context_property_prevent_login_dialog_use_configure.patch)

I believe this patch is neccessary. 
1. HostDictionary needs to parse it - otherwise it's not visible
2. S3Session was not looking at/using the same value - fixed that inconsistency
3. AWSSessionCredentialsRetriever uses url property directly to fetch credentials to workaround issue with Hostparser
4. S3Protocol by-passes login dialogue when context is set (this one I'm not 100% sure if done correctly)

This change would help us a lot - it allows for a full integration of Cyberduck into some publishers' very manual processes when they deliver or retrieve content from Castlabs' software.

dkocher:
> [look_at_preferences_all_the_time_parse_context_property_prevent_login_dialog_use_configure.patch](https://github.com/user-attachments/files/23357400/look_at_preferences_all_the_time_parse_context_property_prevent_login_dialog_use_configure.patch)
> 
> I believe this patch is neccessary.
> 
> 1. HostDictionary needs to parse it - otherwise it's not visible
> 2. S3Session was not looking at/using the same value - fixed that inconsistency
> 3. AWSSessionCredentialsRetriever uses url property directly to fetch credentials to workaround issue with Hostparser
> 4. S3Protocol by-passes login dialogue when context is set (this one I'm not 100% sure if done correctly)
> 
> This change would help us a lot - it allows for a full integration of Cyberduck into some publishers' very manual processes when they deliver or retrieve content from Castlabs' software.

Please open as a pull request to allow it to be commented.

ylangisc (approved):

Issue #13936, opened 2022-12-04
Passwords for ProxyJump hosts in .ssh/config are not saved 


dirkpetersen:
**Is your feature request related to a problem? Please describe.**

When I use the ProxyJump feature the password for the jumphost is never saved (Version 8.5.1 / Windows 11). This is especially problematic since Cyberduck seems to need to re-authenticate each time I hit the upload button so if the files I upload are scattered over multiple folders I have to enter my password 4-5 times which seems a bit hacky

```
Host jumper
    User user1
    Hostname host1.domain.org

Host *
    ProxyJump jumper
```

**Describe the solution you'd like**
save password for jumphost if checked when prompted 

**Describe alternatives you've considered**
Considered ssh keys but are not allowed here

dkocher:
Should have been resolved with 3e28ad3b but apparently we fail to lookup the saved credentials.

dkocher:
> Awesome, thanks a lot !

It would be great if you can verify the fix with build 8.5.2.38764 available through _Preferences… → Update → Snapshot Builds_.

dirkpetersen:
I tested it and it works very well. Thank you!

PR #17659, opened 2025-11-18
Fail when reading private key for selected certificate fails.

### LINKED ISSUE #17657: Display failure and abort connect attempt when private key for selected certificate cannot be retrieved
When the private key for a certificate selected for a connection cannot be retrieved the connect attempt should be aborted with a descriptive error message. The use case is when the private key for a certificate is not available in the system store but on a smartcard only.

dkocher:
Fix #17657. Refactor changes for #13936.


ylangisc:
Does not compile.

ylangisc (approved):

PR #17855, opened 2026-02-02
Use display name for sites

dkocher:
Fix #17572.


ylangisc (approved):

AliveDevil (changes requested):
filter() has a validation against getName(), which should be changed to getDisplayName()

AliveDevil (approved):

PR #17880, opened 2026-02-13
Make sure to set Content-Length header for PUT and POST requests with…

ylangisc:
… an empty body.

### DIFF
--- a/jersey/src/main/java/ch/cyberduck/core/jersey/HttpComponentsConnector.java
+++ b/jersey/src/main/java/ch/cyberduck/core/jersey/HttpComponentsConnector.java
@@ -18,6 +18,7 @@ package ch.cyberduck.core.jersey;
 import ch.cyberduck.core.PreferencesUseragentProvider;
 import ch.cyberduck.core.http.HttpMethodReleaseInputStream;
 
+import org.apache.commons.lang3.StringUtils;
 import org.apache.http.Header;
 import org.apache.http.HttpEntity;
 import org.apache.http.HttpHost;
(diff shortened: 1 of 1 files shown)


jersey/src/main/java/ch/cyberduck/core/jersey/HttpComponentsConnector.java
```diff
@@ -65,16 +67,20 @@ public HttpComponentsConnector(final CloseableHttpClient client, final Configura
     @Override
     public ClientResponse apply(final ClientRequest clientRequest) throws ProcessingException {
         final HttpUriRequest request = this.toUriHttpRequest(clientRequest);
+        if(StringUtils.equalsAny(clientRequest.getMethod(), HttpMethod.POST, HttpMethod.PUT)) {
+            if(this.getHttpEntity(clientRequest) == null) {
```
dkocher: Why don't just check `clientRequest.getEntity()`.

dkocher (changes requested):

jersey/src/main/java/ch/cyberduck/core/jersey/HttpComponentsConnector.java
ylangisc: As `this.getHttpEntity(clientRequest)` is the wrapped entity to be used for the http client. Just wanted to make sure that we operate on the same entity assuming I don't know what the wrapper is doing.

dkocher: `getHttpEntity` is private and cannot be overriden.

dkocher (approved):

PR #17882, opened 2026-02-13
Handle 204 response code for patch operation.

dkocher:


AliveDevil:
This prevents throwing the error into users face.
However, Microsoft OneDrive Consumer fails to handle the patch request to `PATCH v1.0/drives/{driveId}/items/{itemId}` (as the recommended way of calling that endpoint), but it works fine with the undocumented way of specifying the target file in `PATCH v1.0/drives/{driveId}/root:/path/to/file`.
Effectively, when receiving an `Optional.empty()` assume that the rename operation failed.

Repro with MSGraph Sdk Java and ticket with Microsoft pending.

AliveDevil (approved):
I'm OK with this.
There were discussions about disabling Move/Timestamp feature for the next release, which would require releasing a new version in the future, when this starts working again.

Issue #13587, opened 2022-07-19
CLI OAuth workflow never finishes as redirect_url is set wrong


soumith:
**Describe the bug**

In the CLI, take a simple workflow such as `duck -l googledrive:/`.
The CLI opens a browser, which does the OAuth workflow.
Once authenticated, the expectation is that you either have the OAuth redirect the CLI.
However, the OAuth redirect_url is set to open the CyberDuck app, and does not respect that the CLI originated the OAuth workflow.

**To Reproduce**
1. Go to command-line
2. Run `duck -l googledrive:/`
3. Try to finish authentication workflow

**Expected behavior**
You expect the workflow to finish.
Either the OAuth takes you back to CLI and CLI finishes auth, or you get instructions on how to manually finish the workflow.

**Desktop (please complete the following information):**
 - macOS 12.4
 - [Also checked] Linux Ubuntu 20.04

dkocher:
This was missed in #13366.

dkocher:
Not sure how we should implement this, as manual copy & paste of authentication code [^1] is deprecated . 

[^1]: Important: The manual copy/paste option, also referred to as an out of band (OOB) redirect method, is DEPRECATED and [will be removed in October 2022](https://developers.googleblog.com/2022/02/making-oauth-flows-safer.html). Review the [OOB migration guide](https://developers.google.com/identity/protocols/oauth2/resources/oob-migration) for instructions on how to migrate to a secure alternative. https://developers.google.com/identity/protocols/oauth2/native-app#manual-copypaste

soumith:
from the migration guide, the most matching option seems to be a loopback IP address -- so cyberduck temporarily listens on some localhost+port and that is given as the redirect_uri

https://developers.google.com/identity/protocols/oauth2/native-app#redirect-uri_loopback

dkocher:
Relates to #13476.

dkocher:
> from the migration guide, the most matching option seems to be a loopback IP address -- so cyberduck temporarily listens on some localhost+port and that is given as the redirect_uri
> 
> https://developers.google.com/identity/protocols/oauth2/native-app#redirect-uri_loopback

We don't currently have an implementation using a listener on the loopback address as it is prone to man in the middle attacks. it is still feasible for desktop apps according to the documentation [^1]
> …but will continue to be supported on desktop apps.

[^1]: https://developers.google.com/identity/protocols/oauth2/resources/loopback-migration

dkocher:
[...]

PR #17899, opened 2026-02-22
AWS IAM Identity Center Support

### LINKED ISSUE #13377: aws sso based bookmarks are too difficult to use

### LINKED ISSUE #17596: Support iOS URL scheme as redirect URI for OAuth with Google
Temporarily allowed iOS URL scheme using _Bundle ID_ is deprecated.

https://developers.google.com/identity/protocols/oauth2/native-app#installed_app_redirect_methods
> com.googleusercontent.apps.123 is the reverse DNS notation of the client ID.

dkocher:
Resolves #13377. 

- [x] Support for Loopback IP address as OAuth Redirect URI. Resolves #17596. Allows to register `http://localhost/` for CLI connection profiles without requiring users to enter the authorization code manually resolving #13587.


s3/src/main/java/ch/cyberduck/core/sso/RegisterClientOAuth2RequestInterceptor.java
```diff
@@ -0,0 +1,196 @@
...
+    @Override
+    protected IdTokenResponse exchangeToken(final AuthorizationCodeFlow flow, final String authorizationCode) throws BackgroundException {
+        final AWSSSOOIDCClientBuilder configuration = AWSSSOOIDCClientBuilder.standard()
+                .withRegion(region)
+                .withClientConfiguration(new CustomClientConfiguration(host,
+                        new ThreadLocalHostnameDelegatingTrustManager(trust, String.format("oidc.%s.amazonaws.com", region)), key));
+        final AWSSSOOIDC client = configuration.build();
+        try {
+            // Use reflection to access the pkce field from AuthorizationCodeFlow
+            final Field pkceField = AuthorizationCodeFlow.class.getDeclaredField("pkce");
```
ylangisc, line 201: When using reflection we should at least have a test case that makes sure the field names have not been changed and can be accessed.

ylangisc (approved):

PR #17914, opened 2026-03-03
Remove Bulk feature usage

dkocher:


ctera/src/main/java/ch/cyberduck/core/ctera/CteraDirectIOReadFeature.java
```diff
@@ -16,96 +16,137 @@
...
+        catch(IOException e) {
+            throw new HttpExceptionMappingService().map("Download {0} failed", e, file);
+        }
     }
 
     @Override
     public InputStream read(final Path file, final TransferStatus status, final ConnectionCallback callback) throws BackgroundException {
         try {
-            final EncryptInfo key = new EncryptInfo(status.getParameters().get(CTERA_WRAPPEDKEY), session.getOrCreateAPIKeys().secretKey);
+            final DirectIO metadata = this.getMetadata(file);
```
ylangisc: Metadata must be read once only. Move it to bulk feature for example.

ylangisc (approved):

PR #17918, opened 2026-03-04
Remove /Default symbolic link in Sharepoint Sites

dkocher:
Resolve #17917.


AliveDevil:
3e1898e regression from https://github.com/iterate-ch/cyberduck/pull/17855
[SharepointListService](https://github.com/iterate-ch/cyberduck/blob/b5b75c9432ef3ad4be544f54457a11a256d48336/onedrive/src/main/java/ch/cyberduck/core/onedrive/SharepointListService.java#L65) sets "siteCollection.hostname" as symlink target from root query, which doesn't hold anymore, and the path is now the Displayname, optionally with the UUID attached.
Maybe we should delete the Default-symlink now. Don't want to add more complexity to the default-path resolution.

ylangisc (approved):

PR #17933, opened 2026-03-10
Review client authentication.

dkocher:


oauth/src/main/java/ch/cyberduck/core/oauth/OAuth2AuthorizationService.java
```diff
@@ -231,7 +231,7 @@ private IdTokenResponse authorizeWithCode(final LoginCallback prompt) throws Bac
                 method,
                 transport, json,
                 new GenericUrl(tokenServerUrl),
-                new ClientParametersAuthentication(clientid, StringUtils.isNotBlank(clientsecret) ? clientsecret : null),
+                new ClientParametersAuthentication(clientid, clientsecret),
```
ylangisc: Check introduced in `0794fab9` - not necessary anymore?

ylangisc (approved):

PR #17982, opened 2026-04-07
Disable container tests

dkocher:


ylangisc (commented):
Integrationtests are still running on Jenkins.

PR #17999, opened 2026-04-18
osx-folder-recv-complate-updateprogress

james-tech:
bugfix/osx-folder-recv-complate-updateprogress

version 9.4.1 has a bug:
when ftp recv folder with files from remote, after recv all done, the recved folder progress wont update to done status in fider views.

### DIFF
--- a/core/src/main/java/ch/cyberduck/core/worker/AbstractTransferWorker.java
+++ b/core/src/main/java/ch/cyberduck/core/worker/AbstractTransferWorker.java
@@ -396,16 +396,12 @@ public abstract class AbstractTransferWorker extends TransferWorker<Boolean> {
                             icon = IconService.disabled;
                         }
                         final BytecountStreamListener counter = new TransferStreamListener(transfer,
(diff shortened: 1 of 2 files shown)


core/src/main/java/ch/cyberduck/core/worker/AbstractTransferWorker.java
```diff
@@ -471,6 +467,9 @@ private void transferSegment(final TransferStatus segment) throws BackgroundExce
                             }
                         }
                         finally {
+                            // Unpublish Finder NSProgress for this segment; do not wait for the whole transfer (multi-file
```
dkocher, line 470: I would say it is intended that progress stays active on the folder until all files within have been transferred.

dkocher (commented):
I fail to understand why the differentiation between transfer type is required. The listener is called in `StreamCopier` for both sent and received bytes which will be equal unless there is a transfer failure. For downloads progress will be reported when written to disk, for uploads when sent over wire.

PR #18011, opened 2026-04-23
Drop displayname support.

### LINKED ISSUE #18010: Unable to preview files in vault with QuickLook
The filename of the temporary file is a `.c9r` instead of the extension of the decrypted filename.

dkocher:
This reverts commit 188f6a77. Fix #18010.


ylangisc:
188f6a77 was a fix for MD-21641 which will lead to a regression then?

dkocher:
> [188f6a7](https://github.com/iterate-ch/cyberduck/commit/188f6a77b16b2a4b74a3089419fb9cf8140f08d3) was a fix for [MD-21641](https://jira.iterate.ch/browse/MD-21641) which will lead to a regression then?

Good point. We can keep the temporary filename handling thus keeping the fix from #16163 and just include da057e12a505f4e708fce0863020e70daa80eca0 that results in encrypted filenames no longer being used.

ylangisc (approved):

PR #18021, opened 2026-04-25
S3: Disable versioned listing by default, add preference toggle

### LINKED ISSUE #18020: Endless spinner when opening a large s3 bucket with tons of versioning
**Describe the bug**
When trying to connect to an s3 bucket that has versioning enabled, cyberduck attempts to grab all the files and their historical versions using the S3 List Object Versions API instead of the regular List Objects API. 

That API returns every historical version of every file, so instead of getting back "here are 5 items at the root", S3 was returning thousands of records, paginated 1,000 at a time. Cyberduck had to exhaust all its pages before it could display anything.

**To Reproduce**
1. Go to an s3 bucket with versioning enabled
2. Have the files in there change tons of times (100k+ versions)
3. Try opening the bucket

**Expected behavior**
The default should be to call the regular object list api instead of versioning.
That config already exists in /cyberduck-master/defaults/src/main/resources/default.properties but is not exposed in the gui and is defaulted to true.

 - OS: macOS
 - Version 9.4.1

**Additional context**

ayounes2:
S3: Disable versioned listing by default, add preference toggle

Listing directories in versioned buckets where a file has accumulated
a large number of historical versions (e.g. 100k+) caused the browser
to hang indefinitely. S3VersionedObjectListService paginates through
every historical version before rendering anything.

Change s3.listing.versioning.enable default to false so the standard
non-versioned listing is used by default. Add a "Versioning" checkbox
to S3 Preferences so users can re-enable it when needed.

Fixes #18020

Screenshot 2026-04-25 at 11 45 20 AM

### DIFF
--- a/defaults/src/main/resources/default.properties
+++ b/defaults/src/main/resources/default.properties
@@ -322,7 +322,7 @@ s3.url.expire.seconds=86400
 s3.listing.chunksize=1000
 s3.listing.concurrency=25
 # Allow to disable versioning aware list service regardless of bucket versioning status
-s3.listing.versioning.enable=true
+s3.listing.versioning.enable=false
 
 # Read metadata of every file in list service to display modification date stored in metadata
 s3.listing.metadata.enable=false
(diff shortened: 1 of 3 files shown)


dkocher:
Thanks a lot for contribution. I have amended this in #18031.

PR #18031, opened 2026-04-28
S3: Disable versioned listing by default, add preference toggle

### LINKED ISSUE #18020: Endless spinner when opening a large s3 bucket with tons of versioning

dkocher:
Amended #18021 with localizations and implementation for Windows.


ylangisc:
UI needs to be fixed.

image

ylangisc (approved):

PR #18331, opened 2026-07-30
Add AWS Console browser sign-in for S3

whatever60:
## Summary

Add a bundled **Amazon S3 (AWS Console Sign-In)** profile with native browser authentication. It requires neither AWS CLI nor long-lived access keys.

- Run the AWS Sign-In authorization-code flow with PKCE and DPoP.
- Keep short-lived S3 access credentials in memory only.
- Persist only the rotating refresh token and DPoP private key through Cyberduck's existing secure password store.
- Refresh credentials automatically for the AWS browser-session lifetime, then reopen the browser when sign-in is required again.
- Display and pin the selected AWS account and IAM user or role to the bookmark.
- Return a successful loopback callback page after browser authentication.

## Motivation

The existing S3 profile expects access keys, while the IAM Identity Center flow requires organization-side Identity Center configuration. AWS Console Sign-In lets a user authenticate with an existing AWS Console identity without creating persistent access keys, installing AWS CLI, or having IAM Identity Center access.

## Implementation notes

- A dedicated S3 profile selects the native sign-in credential strategy without changing existing S3 authentication modes.
- Authorization uses an ephemeral loopback port bound to `127.0.0.1`, exact state validation, PKCE S256, and ES256 DPoP.
- The loopback listener is bound before the authorization URL is generated, avoiding a close/rebind race.
- AWS access key ID, secret access key, and session token are never written to a bookmark or password store.
- The initial ID token pins the AWS account and normalized IAM user or role. A later sign-in with another identity is rejected for that bookmark.
- The callback response is completed before the application resumes, so the browser shows **Login successful** instead of a refused loopback connection.

## Validation

- `mvn -pl oauth,s3 -am -DskipTests package`: all 8 reactor modules passed on macOS and Windows.
- Focused native sign-in tests: 3 passed on macOS and Windows.
- Broader OAuth/S3 reactor suite: 49 S3 tests passed on macOS; 52 S3 tests passed in an elevated Windows 11 ARM64 VM. Three Docker-only integration tests were excluded.
- Packaged macOS and Windows GUI smoke tests: native browser sign-in, identity pinning, live listing of 59 S3 buckets, full app exit/relaunch, and reconnect without reopening the browser all passed.
- Password-store audit confirmed that temporary AWS access credentials are not persisted; only the refresh token and DPoP key are retained using Cyberduck's standard secure-storage path.
- `git diff --check origin/master...HEAD` passed.

### DIFF
--- a/core/src/main/java/ch/cyberduck/core/LoginOptions.java
+++ b/core/src/main/java/ch/cyberduck/core/LoginOptions.java
@@ -100,8 +100,7 @@ public final class LoginOptions {
         icon = protocol.disk();
         usernamePlaceholder = protocol.getUsernamePlaceholder();
         passwordPlaceholder = protocol.getPasswordPlaceholder();
-        return this.keychain(Boolean.parseBoolean(protocol.getProperties().getOrDefault(
-                "connection.login.keychain", Boolean.TRUE.toString())));
+        return this;
     }
 
     public LoginOptions user(boolean e) {
--- a/core/src/main/java/ch/cyberduck/core/LoginOptions.java
+++ b/core/src/main/java/ch/cyberduck/core/LoginOptions.java
@@ -100,6 +100,8 @@ public final class LoginOptions {
         icon = protocol.disk();
         usernamePlaceholder = protocol.getUsernamePlaceholder();
         passwordPlaceholder = protocol.getPasswordPlaceholder();
+        keychain = password || token || oauth || publickey || certificate;
+        save = keychain && PreferencesFactory.get().getBoolean("connection.login.keychain");
         return this;
     }
 
--- a/core/src/main/java/ch/cyberduck/core/LoginOptions.java
+++ b/core/src/main/java/ch/cyberduck/core/LoginOptions.java
@@ -100,9 +100,8 @@ public final class LoginOptions {
         icon = protocol.disk();
         usernamePlaceholder = protocol.getUsernamePlaceholder();
         passwordPlaceholder = protocol.getPasswordPlaceholder();
-        keychain = password || token || oauth || publickey || certificate;
-        save = keychain && PreferencesFactory.get().getBoolean("connection.login.keychain");
-        return this;
+        return this.keychain(Boolean.parseBoolean(protocol.getProperties().getOrDefault(
+                "connection.login.keychain", Boolean.TRUE.toString())));
     }
 
     public LoginOptions user(boolean e) {
--- a/oauth/src/main/java/ch/cyberduck/core/oauth/LoopbackOAuth2AuthorizationCodeProvider.java
+++ b/oauth/src/main/java/ch/cyberduck/core/oauth/LoopbackOAuth2AuthorizationCodeProvider.java
@@ -34,17 +34,17 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
 import java.io.IOException;
+import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.URI;
-import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
 import java.util.List;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Function;
 
-import com.sun.net.httpserver.HttpExchange;
-import com.sun.net.httpserver.HttpHandler;
 import com.sun.net.httpserver.HttpServer;
 
 public class LoopbackOAuth2AuthorizationCodeProvider extends BrowserOAuth2AuthorizationCodeProvider {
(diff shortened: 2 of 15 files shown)


osx/build.xml
```diff
@@ -31,7 +31,7 @@
...
+              value="-client --add-opens=java.base/sun.security.ssl=ALL-UNNAMED --add-opens=java.base/sun.security.util=ALL-UNNAMED -Djdk.lang.Process.launchMechanism=FORK -Djava.library.path=$APP_PACKAGE/Contents/Frameworks -Djna.boot.library.path=$APP_PACKAGE/Contents/Frameworks -Djna.library.path=$APP_PACKAGE/Contents/Frameworks -Djna.nounpack=true -Djava.awt.headless=true -Dsun.jnu.encoding=utf-8 -Dfile.encoding=utf-8 -Dsun.io.useCanonCaches=false -DLog4jContextSelector=org.apache.logging.log4j.core.selector.BasicContextSelector -XX:+UseShenandoahGC -XX:ShenandoahGCHeuristics=compact -XX:MinHeapFreeRatio=10 -XX:MaxHeapFreeRatio=20 -XX:+UseStringDeduplication"/>
```
dkocher: With the recent fix in #18320 this should not be required.

s3/src/main/java/ch/cyberduck/core/s3/S3Session.java
```diff
@@ -265,7 +266,12 @@ public void process(final HttpRequest request, final HttpContext context) {
     }
 
     protected S3CredentialsStrategy configureCredentialsStrategy(final HttpClientBuilder configuration,
-                                                                 final LoginCallback prompt) throws BackgroundException {
+                                                                 final LoginCallback prompt,
+                                                                 final CancelCallback cancel) throws BackgroundException {
+        if(S3LoginProtocol.IDENTIFIER.equals(host.getProtocol().getIdentifier())) {
+            log.debug("Configure AWS Console Sign-In");
+            return new AWSConsoleLoginCredentialsStrategy(host, cancel);
+        }
```
dkocher: Will need to abstract this to determine the logging strategy without referencing concrete protocol subclass.

s3/src/main/java/ch/cyberduck/core/s3/S3Session.java
```diff
@@ -360,7 +366,9 @@ public void login(final LoginCallback prompt, final CancelCallback cancel) throw
                 if(!credentials.isAnonymousLogin()) {
                     // Returns details about the IAM user or role whose credentials are used to call the operation.
                     // No permissions are required to perform this operation.
-                    new STSAuthorizationService(host, trust, key, prompt).getCallerIdentity(credentials);
+                    if(!S3LoginProtocol.IDENTIFIER.equals(host.getProtocol().getIdentifier())) {
+                        new STSAuthorizationService(host, trust, key, prompt).getCallerIdentity(credentials);
+                    }
```
dkocher: Will need to abstract this to determine the logging strategy without referencing concrete protocol subclass.

windows/src/main/csharp/ch/cyberduck/ui/controller/BookmarkController.cs
```diff
@@ -617,6 +617,7 @@ private void View_ChangedProtocolEvent()
                 _host.setPort(HostnameConfiguratorFactory.get(selected).getPort(_host.getHostname()));
                 _host.setCredentials(CredentialsConfiguratorFactory.get(selected).configure(_host));
                 _options.configure(selected);
+                _host.getCredentials().setSaved(_options.keychain() && _host.getCredentials().isSaved());
```
dkocher: We can change the behaviour by setting the property `connection.login.keychain` in the connection profile instead making this change obsolete..

osx/src/main/java/ch/cyberduck/ui/cocoa/controller/BookmarkController.java
```diff
@@ -294,6 +294,7 @@ public void protocolSelectionChanged(final NSPopUpButton sender) {
             bookmark.setPort(HostnameConfiguratorFactory.get(selected).getPort(bookmark.getHostname()));
             bookmark.setCredentials(CredentialsConfiguratorFactory.get(selected).configure(bookmark));
             options.configure(selected);
+            bookmark.getCredentials().setSaved(options.keychain && bookmark.getCredentials().isSaved());
```
dkocher: We can change the behaviour by setting the property `connection.login.keychain` in the connection profile instead making this change obsolete..

core/src/main/java/ch/cyberduck/core/LoginOptions.java
```diff
@@ -100,6 +100,8 @@ public LoginOptions configure(final Protocol protocol) {
         icon = protocol.disk();
         usernamePlaceholder = protocol.getUsernamePlaceholder();
         passwordPlaceholder = protocol.getPasswordPlaceholder();
+        keychain = password || token || oauth || publickey || certificate;
```
dkocher: Not sure why this additional checks are required.

dkocher (changes requested):
Thanks for your submission! Not sure we want to have this integration relying on command line tool invocations generally. We worked to remove the dependency on AWS CLI with #13377.

s3/src/main/java/ch/cyberduck/core/signin/AWSConsoleLoginCredentialsStrategy.java
```diff
@@ -0,0 +1,314 @@
...
+        final String code = new LoopbackOAuth2AuthorizationCodeProvider().prompt(host, prompt,
+                        redirect -> {
+                            redirectUri.set(redirect);
+                            return new URIBuilder(URI.create(String.format("%s/v1/authorize", endpoint)))
+                                .addParameter("response_type", "code")
+                                .addParameter("client_id", CLIENT_ID)
+                                .addParameter("state", state)
+                                .addParameter("code_challenge", challenge)
+                                .addParameter("code_challenge_method", "SHA-256")
+                                .addParameter("scope", "openid")
```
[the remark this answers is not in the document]
whatever60: AWS Sign-In intentionally differs from the RFC 7636 spelling here: its `AuthorizeOAuth2Access` API requires the literal `SHA-256` value for `codeChallengeMethod`, so the current code follows the AWS service contract. See https://docs.aws.amazon.com/signin/latest/APIReference/API_dataplane-signin_AuthorizeOAuth2Access.html

whatever60:
Hi David, thank you for the detailed review. I now have a better understanding of this repo and its design principles. I agree that removing CLI is the right call that makes feature impls more consistent and self-contained.

I have removed the direct AWS CLI invocation and replaced it with native AWS auth flow using PKCE and DPoP, by reusing Cyberduck's existing components like loopback OAuth, S3 credential strategy, connection profile, and secure pw store.

I also removed the unrelated UI, keychain, and process launch changes, and now select the auth strategy through a connection profile property rather than concrete protocol class checks.

This PR is ready for another review. Thanks again and I would appreciate another look.

Yiming

oauth/src/main/java/ch/cyberduck/core/oauth/LoopbackOAuth2AuthorizationCodeProvider.java
```diff
@@ -34,77 +34,97 @@
...
-        registry.register(state, new OAuth2TokenListener() {
-            @Override
-            public void callback(final String code) {
-                log.info("Callback with code {}", code);
-                if(!StringUtils.isBlank(code)) {
-                    authenticationCode.set(code);
-                }
+        OAuth2TokenListenerRegistry.get().register(expectedState, code -> {
+            if(StringUtils.isBlank(code)) {
                 signal.countDown();
```
dkocher: Not always signaling looks like a regression.

oauth/src/main/java/ch/cyberduck/core/oauth/LoopbackOAuth2AuthorizationCodeProvider.java
```diff
@@ -34,77 +34,97 @@
...
+            else {
+                authenticationCode.set(code);
+            }
         });
         try {
-            final HttpServer server = HttpServer.create(new InetSocketAddress(
-                    URI.create(redirectUri).getHost(), -1 == URI.create(redirectUri).getPort() ? 0 : URI.create(redirectUri).getPort()), 0);
+            final URI requested = null == requestedRedirectUri ? null : URI.create(requestedRedirectUri);
+            final HttpServer server = HttpServer.create(null == requested ?
+                    new InetSocketAddress(InetAddress.getByAddress(new byte[]{127, 0, 0, 1}), 0) :
```
dkocher: Why not just use `InetAddress.getLoopbackAddress()`?

s3/src/main/java/ch/cyberduck/core/signin/AWSConsoleLoginCredentialsStrategy.java
```diff
@@ -0,0 +1,314 @@
...
+import com.nimbusds.jose.jwk.Curve;
+import com.nimbusds.jose.jwk.ECKey;
+import com.nimbusds.jose.jwk.gen.ECKeyGenerator;
+import com.nimbusds.jwt.JWTClaimsSet;
+import com.nimbusds.jwt.SignedJWT;
+
+public class AWSConsoleLoginCredentialsStrategy implements S3CredentialsStrategy {
+    private static final Logger log = LogManager.getLogger(AWSConsoleLoginCredentialsStrategy.class);
+    private static final ObjectMapper MAPPER = new ObjectMapper();
+    private static final SecureRandom RANDOM = new SecureRandom();
```
dkocher: Make ivar or local scope.

PR #18362, opened 2026-08-13
Resolve default path of bookmark with delimiter in file name (Google Drive)

gsamat:
Fixes #17931.

### Problem

A bookmark pointing into a Shared Drive whose name contains `/` mounts the Google Drive root instead of the bookmarked folder.

Bookmarks persist the working directory as a single string in `defaultpath`. In Google Drive `/` is both the path delimiter and a legal character in file and Shared Drive names, so that string is ambiguous. For the bookmark in #17931, with a Shared Drive named `Либо/Либо `, the default path

```
/Shared Drives/Либо/Либо /Подкасты/Запуск завтра/14 сезон (видео!)
```

is split into the segments `Либо` and `Либо `. No Shared Drive by that name exists, `PathAttributesHomeFeature` fails with `NotfoundException detail='/Shared Drives/Либо'`, and `MountWorker` falls back to mounting `/` — which is what the user sees.

### Change

`DriveDefaultPathHomeFeature` resolves the default path of a bookmark and is registered as the `Home` feature in `DriveSession`.

It first composes the path exactly as `DefaultPathHomeFeature` does today, so behaviour and request count are unchanged for the common case. Only when that fails with `NotfoundException` does it match the segments of the default path against the file names on the server, joining consecutive segments with the delimiter and backtracking when a shorter match turns out to have no matching children (both a drive `a` and a drive `a/b` may exist).

### This is not a pretty solution, and I know it

Recovering the segmentation from an ambiguous string is guesswork, so the fallback is a heuristic:

* **Ambiguity is resolved arbitrarily.** If more than one reading of the path exists on the server, the shortest matching one wins. There is no correct answer — the information was lost when the bookmark was saved.
* **Not everything is recoverable.** Names containing `//` or a trailing delimiter do not survive `PathNormalizer.normalize` before this code runs.
* **No cap on the search.** In the pathological case where nearly every prefix matches an existing name, the number of lookups grows exponentially with the number of segments. In practice it is one lookup per level, and the search only runs after the fast path has already failed.

I went for it anyway because Google Drive bookmarks silently breaking is worse than an imperfect heuristic in a code path that today has no chance of succeeding at all. Consider it a stopgap.

### The proper fix, for what it is worth

The bookmark format can already express this without ambiguity. `Host` serializes the working directory as an object under `Workdir Dictionary` (`Host.java`, `HostDictionary.java`), where `Path` keeps `Name` and `Parent` as separate fields and carries `File Id` in its attributes — a name containing the delimiter round-trips fine. `WorkdirHomeFeature` is already first in the `Home` chain, so a persisted working directory would be used before anything in this PR. It is only cleared for bookmarks, in `BookmarkCollection#collectionItemAdded`.

Persisting it for bookmarks touches decisions I did not want to make in a bugfix: editing the path in the bookmark UI would have to invalidate the stored identifier, exported bookmarks and `.duck` files shared between accounts need a meaningful fallback, and there is a policy question about a file that was moved (same id, different path) versus deleted and recreated (same path, different id) — plus the same core is used by the CLI and Mountain Duck. Happy to look into it if you think it is the right direction.

Note this fallback would still be needed either way: default paths that arrive as a plain string — existing bookmarks, URLs, the command line, imports from other clients — remain ambiguous.

### Testing

`DriveDefaultPathHomeFeatureTest` covers the plain case, a Shared Drive name containing the delimiter, backtracking on an ambiguous match, not found, no default path and the root default path, against a stubbed `AttributesFinder`. Removing the fallback makes the two delimiter tests fail with the same `NotfoundException` as reported in the issue.

I could not run the integration tests of the module — no test account. I did verify the fix end to end against the actual Google Drive account from the issue, with a build of this branch:

```
WARN  DriveDefaultPathHomeFeature - Failure NotfoundException detail='/Shared Drives/Либо'
      finding default path .../14 сезон (видео!). Match path segments against file names
INFO  MountWorker - Mount path Path{path='/Shared Drives/Либо/Либо /Подкасты/Запуск завтра/14 сезон (видео!)'}
INFO  Retrieved chunk of 54 items in .../14 сезон (видео!)
```

Before the change the same bookmark logged `Fallback to mount path Path{path='/'}`.

### DIFF
--- a/googledrive/src/main/java/ch/cyberduck/core/googledrive/DriveDefaultPathHomeFeature.java
+++ b/googledrive/src/main/java/ch/cyberduck/core/googledrive/DriveDefaultPathHomeFeature.java
@@ -1,0 +1,108 @@
+package ch.cyberduck.core.googledrive;
+
+/*
+ * Copyright (c) 2002-2026 iterate GmbH. All rights reserved.
+ * https://cyberduck.io/
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ */
+
+import ch.cyberduck.core.DefaultPathAttributes;
+import ch.cyberduck.core.Host;
+import ch.cyberduck.core.Path;
+import ch.cyberduck.core.PathAttributes;
+import ch.cyberduck.core.PathNormalizer;
+import ch.cyberduck.core.exception.BackgroundException;
+import ch.cyberduck.core.exception.NotfoundException;
+import ch.cyberduck.core.features.AttributesFinder;
+import ch.cyberduck.core.features.Home;
+import ch.cyberduck.core.shared.AbstractHomeFeature;
+
+import org.apache.commons.lang3.StringUtils;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import java.util.EnumSet;
+
+/**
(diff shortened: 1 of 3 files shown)


dkocher:
We will want a generic way to add support for slash in any filename and all supported protocols.

PR #18393, opened 2026-08-28
Handle timeout during resumable file upload finalization by confirming offset.

dkocher:


tus/src/main/java/ch/cyberduck/core/tus/TusWriteFeature.java
```diff
@@ -74,6 +81,15 @@ public Void call(final HttpEntity entity) throws BackgroundException {
                 request.setHeader(TUS_HEADER_UPLOAD_OFFSET, String.valueOf(status.getOffset()));
                 // All PATCH requests MUST use Content-Type: application/offset+octet-stream
                 request.setHeader(HttpHeaders.CONTENT_TYPE, "application/offset+octet-stream");
+                // Last chunk completing the upload. The server may take a significant amount of time to respond
+                // while assembling and validating previously uploaded chunks
+                final boolean finalize = (status.getOffset() + status.getLength()) == status.getParent().getLength();
+                if(finalize) {
```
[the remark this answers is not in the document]
dkocher: When wrapping with `CryptoTransferStatus` all values are copied including `parent`. The compared length are expected to be the encrypted size in case of the features being wrapped by `CryptoWriteFeature` and `CryptoUploadFeature`.

cryptomator/src/main/java/ch/cyberduck/core/cryptomator/CryptoTransferStatus.java
```diff
@@ -41,6 +41,10 @@ public CryptoTransferStatus(final Vault vault, final Path file, final TransferSt
                 // Assume single chunk upload
                 .setOffset(0L == proxy.getOffset() ? 0L : vault.toCiphertextSize(0L, proxy.getOffset()))
                 .setMime(null);
+        // Retrieve plaintext overall status
+        final TransferStatus plaintext = proxy.getParent();
+        // Set ciphertext overall status
+        this.setParent(plaintext.getParent());
```
ylangisc, line 47: **Regression: vault writes on EUE now declare the plaintext size**

When `CryptoWriteFeature` wraps a status that has no parent, `proxy.getParent()` is the proxy itself, so `plaintext.getParent()` is again the proxy and the ciphertext status ends up with `parent == plaintext`. The same happens for the overall status linked by `CryptoUploadFeature`: `new CryptoTransferStatus(vault, file, status)` is evaluated before `setParent` runs, so `cipher.parent = plaintext` and the two form a cycle. Wrapping that status again in `CryptoWriteFeature` yields `parent = cipherA.getParent() = plaintext`.

`EueUploadHelper.createResource/updateResource` now use `status.getParent().getLength()` and therefore send the plaintext size where master sent the ciphertext size:
- Touch in a vault: `CryptoTouchFeature` → `DefaultTouchFeature` (length 0) → `EueMultipartWriteFeature.write` → `createResource` with `size=0` while the 68-byte header is written.
- Upload ≥ `eue.upload.multipart.threshold`: `EueUploadService` → `HttpUploadFeature.transfer` → `CryptoWriteFeature` → `EueWriteFeature.write` (`url == null`) → `createResource` with the plaintext length while ~10.5 MiB of ciphertext is uploaded.
- Copy within a vault via `DefaultCopyFeature`/`MultipartWrite` regresses the same way.

`CryptoTransferStatusTest` only covers the segment hop, where the arithmetic happens to work.

core/src/main/java/ch/cyberduck/core/transfer/TransferStatus.java
```diff
@@ -558,6 +545,18 @@ public TransferStatus setUrl(final String url) {
         return this;
     }
 
+    public TransferStatus setParent(TransferStatus parent) {
+        this.parent = parent;
+        return this;
+    }
+
+    public TransferStatus getParent() {
```
ylangisc, line 553: **Design: `getParent()` has receiver-dependent meaning and a self fallback**

`destinationlength` had one meaning with a defined fallback and was copied through `CryptoTransferStatus` automatically. `parent` yields different things depending on the receiver: segment → plaintext overall, plaintext overall → ciphertext overall, ciphertext overall → plaintext overall, unlinked → `this`. Readers in `BoxUploadHelper`, `EueUploadHelper`, `TusUploadFeature.start` and `TusWriteFeature` all want "total bytes the server will receive" but only get it if the hop count lines up (see the EUE comment on `CryptoTransferStatus`).

Every segment producer also has to hand-wire `.setParent(overall)`. `S3MultipartWriteFeature`, `BrickMultipartWriteFeature` and `BrickTouchFeature` still call `setSegment(true)` without a parent; harmless today, but any write feature adopting the `offset + length == getParent().getLength()` idiom would treat every chunk as final there.

A dedicated accessor for the destination length, set explicitly by `CryptoTransferStatus` from `vault.toCiphertextSize(0, overall.getLength())`, would remove the hop-count convention. Note `HttpUploadFeature.upload` receives `overall` only as `StreamCancelation`, so centralizing the link needs a signature change.

tus/src/main/java/ch/cyberduck/core/tus/TusWriteFeature.java
```diff
@@ -74,8 +77,17 @@ public Void call(final HttpEntity entity) throws BackgroundException {
                 request.setHeader(TUS_HEADER_UPLOAD_OFFSET, String.valueOf(status.getOffset()));
                 // All PATCH requests MUST use Content-Type: application/offset+octet-stream
                 request.setHeader(HttpHeaders.CONTENT_TYPE, "application/offset+octet-stream");
+                // Last chunk completing the upload. The server may take a significant amount of time to respond
+                // while assembling and validating previously uploaded chunks
+                final boolean finalize = (status.getOffset() + status.getLength()) == status.getParent().getLength();
```
ylangisc, line 82: **`finalize` is never true on resumed uploads**

`ResumeFilter.prepare` sets `status.setLength(total - offset)` and `status.setOffset(offset)` on the overall status, while `TusUploadFeature.upload` creates segments with absolute offsets. For the real last chunk `offset + length == total`, but `getParent().getLength() == total - offset`, so `finalize` is false on any resume and the extended timeout plus offset confirmation never engage on exactly the retry path this PR targets. Example with the 10 MiB chunk size: 30 MiB file resumed at 10 MiB → overall length 20 MiB; segment [20,30) gives 30 ≠ 20 (not detected), segment [10,20) gives 20 == 20 (intermediate chunk misdetected). The vault case behaves the same since `CryptoTransferStatus` uses `toCiphertextSize(offset, remaining)`.

Line 105 has the same unit mismatch, comparing the remaining length against the server's absolute `Upload-Offset`. Comparing against `getParent().getOffset() + getParent().getLength()` in both places works for fresh and resumed uploads.

tus/src/main/java/ch/cyberduck/core/tus/TusWriteFeature.java
```diff
@@ -74,8 +77,17 @@ public Void call(final HttpEntity entity) throws BackgroundException {
...
+                if(finalize) {
+                    final RequestConfig context = session.getClient().getContext().getRequestConfig();
```
ylangisc, line 84: **`getContext().getRequestConfig()` returns `RequestConfig.DEFAULT`, not the client configuration**

`DAVClient.execute` creates a fresh `HttpClientContext` per request and never sets `REQUEST_CONFIG` on the persistent `SardineImpl.context` that `getContext()` returns (`DAVSession.alert()` only touches it for insecure hosts, and also with defaults). `HttpClientContext.getRequestConfig()` therefore falls back to `RequestConfig.DEFAULT`, and a request-level config replaces the client default without merging (`InternalHttpClient.doExecute`).

The finalize PATCH is thus sent with `connectTimeout=-1`, `connectionRequestTimeout=-1` and `normalizeUri=true`, losing `connection.timeout.seconds`, `http.manager.timeout` and `http.request.uri.normalize=false` from `HttpConnectionPoolBuilder.createRequestConfig`. A stalled connect or exhausted pool waits forever, and an upload URL with dot segments or percent-encoding gets rewritten only on the last chunk.

Copying from `((Configurable) session.getClient().getClient()).getConfig()` (the client's default config) and only overriding the socket timeout keeps the pool settings.

ylangisc (commented):
Review of the timeout handling and the `destinationlength` → `parent` refactoring. Three confirmed issues (EUE vault size regression, `finalize` detection on resumed uploads, `RequestConfig` copied from an unpopulated context), one design concern on `getParent()`, two plausible edge cases and two minor consistency points. Details inline.

tus/src/main/java/ch/cyberduck/core/tus/TusWriteFeature.java
```diff
@@ -86,6 +98,22 @@ public Void handleResponse(final HttpResponse response) throws HttpResponseExcep
                         }
                     });
                 }
+                catch(SocketTimeoutException e) {
+                    if(finalize) {
+                        log.warn("Timeout waiting for response completing upload of {} to {}", file, request.getURI(), e);
+                        try {
+                            if(status.getParent().getLength() == new TusUploadHelper(session).offset(request.getURI().toString())) {
```
ylangisc, line 105: **Failed confirmation HEAD leads to a duplicate PATCH and 409** (plausible, not confirmed)

The HEAD runs with the default 30 s socket timeout on a fresh context. If it fails or times out (server still assembling, or the upload resource was removed on completion and HEAD returns 404), the `BackgroundException` is swallowed and the original timeout is rethrown as `ConnectionTimeoutException`. `DefaultFailureDiagnostics` classifies that as network and `DefaultRetryCallable` re-runs the chunk, sending the same PATCH at the same `Upload-Offset`. The server answers 409 → `ConflictException` (no further retry) and the transfer is reported failed although the file is complete.

This narrows a pre-existing failure mode rather than introducing it. Retrying the HEAD with the extended timeout before rethrowing, or verifying the target via `Find`/attributes, would close the gap. Confirming requires observing oCIS HEAD behavior right after a completed upload.

tus/src/main/java/ch/cyberduck/core/tus/TusUploadFeature.java
```diff
@@ -197,40 +196,20 @@ public Void call() throws BackgroundException {
...
-                    // If the resource is not found, the Server SHOULD return either the 404 Not Found,
-                    // 410 Gone or 403 Forbidden status without the Upload-Offset header.
-                    throw new HttpResponseException(response.getStatusLine().getStatusCode(), response.getStatusLine().getReasonPhrase());
-                }
-            });
-            return new Write.Append(true).withStatus(status).withOffset(offset);
+            return new Write.Append(true).withStatus(status).withOffset(new TusUploadHelper(session).offset(uploadUrl));
         }
-        catch(HttpResponseException e) {
+        catch(NotfoundException | AccessDeniedException e) {
```
ylangisc: **Stale upload URL is kept on `InteroperabilityException`** (plausible)

Master reset the stored URL on any `HttpResponseException`. `ConflictException`, `LockedException` and `RetriableAccessDeniedException` are subclasses of `AccessDeniedException`, and 408/502/503/504 are retried as network failures, so those are fine. But anything mapped to `InteroperabilityException` by the default branch (200 without `Upload-Offset`, 204, 400, other unmapped codes) now propagates out of `append()`, the `tus.url.*` preference survives, and every later transfer of that file repeats the same HEAD and fails identically until preferences are cleared.

Adding `InteroperabilityException` to the catch (the pattern used in `AbstractUploadFilter`) would keep the new behavior of not discarding the URL on network or credential errors while avoiding the sticky failure. Trigger needs a non-spec-compliant server response.

box/src/main/java/ch/cyberduck/core/box/BoxUploadHelper.java
```diff
@@ -83,8 +83,8 @@ public UploadSession createUploadSession(final TransferStatus status, final Path
                 final FilesUploadSessionsBody uploadSessionsBody = new FilesUploadSessionsBody()
                         .folderId(fileid.getFileId(file.getParent()))
                         .fileName(file.getName());
-                if(status.getLength() != TransferStatus.UNKNOWN_LENGTH) {
-                    uploadSessionsBody.fileSize(status.getLength());
+                if(status.getParent().getLength() != TransferStatus.UNKNOWN_LENGTH) {
```
ylangisc, line 86: **Only the create branch switched to `getParent().getLength()`**

The `isExists()` branch above (line 74) still uses `status.getLength()`. Currently latent because `BoxThresholdUploadService` excludes vault files from `BoxLargeUploadService`, so `parent == this`. If the large service is ever enabled for vaults, overwriting declares the plaintext size while creating declares the ciphertext size. Both branches should agree.

tus/src/main/java/ch/cyberduck/core/tus/TusUploadHelper.java
```diff
@@ -0,0 +1,80 @@
...
+import org.apache.http.client.HttpResponseException;
+import org.apache.http.client.ResponseHandler;
+import org.apache.http.client.methods.HttpHead;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import java.io.IOException;
+
+import static ch.cyberduck.core.tus.TusCapabilities.*;
+import static ch.cyberduck.core.tus.TusCapabilities.TUS_HEADER_UPLOAD_OFFSET;
```
ylangisc: Redundant: already covered by the wildcard static import on the previous line (`TusUploadFeature` and `TusWriteFeature` use only the wildcard). Also the constructor parameter on line 43 and `TransferStatus.setParent` lack `final`, unlike the neighbouring constructors and setters.

tus/src/main/java/ch/cyberduck/core/tus/TusUploadFeature.java
dkocher: In 40eb7365.

box/src/main/java/ch/cyberduck/core/box/BoxUploadHelper.java
dkocher: In 551fd513.

tus/src/main/java/ch/cyberduck/core/tus/TusWriteFeature.java
dkocher: Confirmed. `DAVClient.execute` creates a new `HttpClientContext` per request, so `getContext().getRequestConfig()` always returned `RequestConfig.DEFAULT`. Fixed in 650a18ea6e by copying the default request configuration from the client (`Configurable#getConfig`) and only overriding the socket timeout.

tus/src/main/java/ch/cyberduck/core/tus/TusWriteFeature.java
dkocher: Confirmed. Segments use absolute offsets while the overall status carries the resume offset and remaining length after `ResumeFilter.prepare`. Fixed in e07bdb5b4f by comparing against `getParent().getOffset() + getParent().getLength()` both for detecting the final chunk and when confirming completion by querying the offset after a timeout.

tus/src/main/java/ch/cyberduck/core/tus/TusWriteFeature.java
dkocher: In e07bdb5b4fbabc890ee9c71df811c179fd2e625c.

cryptomator/src/main/java/ch/cyberduck/core/cryptomator/CryptoTransferStatus.java
dkocher: Confirmed. Wrapping a status without parent resolved the parent to the plaintext status, and `CryptoUploadFeature` created a cycle between plaintext and ciphertext overall status. Fixed in 1b08ea4055 by only setting the ciphertext overall status when wrapping a segment (`TransferStatus#isSegment`) and clearing the parent otherwise. Added tests for wrapping a status without parent and the overall status attached in `CryptoUploadFeature`.

core/src/main/java/ch/cyberduck/core/transfer/TransferStatus.java
dkocher: Addressed the concrete issues, keeping the parent link rather than reintroducing a dedicated destination length, since the TUS finalization check also needs the overall offset when resuming:
- Ciphertext overall status no longer resolves to the plaintext status; `CryptoTransferStatus` only links the ciphertext overall status for segments (1b08ea4055).
- `TusWriteFeature` compares against `getParent().getOffset() + getParent().getLength()` (e07bdb5b4f).
- `S3MultipartWriteFeature`, `BrickMultipartWriteFeature` and `AbstractDownloadFilter` set the parent for segments. Removed the unused segment flag in `BrickTouchFeature` (dae37a1e4a).
- Documented the semantics of `TransferStatus#getParent` (38648d93ed).

Centralizing the link in `HttpUploadFeature` requires a signature change and is left for a separate refactor.

ylangisc (approved):

PR #18399, opened 2026-08-31
Revert "Disable TLS1.3 by default for FTP connections due to interope…

dkocher:
…rability issues with many servers."

This reverts commit c0524b78

### DIFF
--- a/defaults/src/main/resources/default.properties
+++ b/defaults/src/main/resources/default.properties
@@ -622,7 +622,6 @@ connection.unsecure.warning.smb=true
 connection.ssl.provider.bouncycastle.position=1
 # Register bouncy castle as preferred provider. Used in Cyptomator, SSL and SSH
 connection.ssl.protocols=TLSv1.3,TLSv1.2,TLSv1.1,TLSv1
-connection.ssl.protocols.ftp=TLSv1.2,TLSv1.1,TLSv1
 connection.ssl.cipher.blacklist=
 connection.ssl.x509.revocation.online=false
 # Default secure random strong algorithm
--- a/ftp/src/main/java/ch/cyberduck/core/ftp/FTPSession.java
+++ b/ftp/src/main/java/ch/cyberduck/core/ftp/FTPSession.java
@@ -144,7 +144,7 @@ public class FTPSession extends SSLSession<FTPClient> {
     protected FTPClient connect(final ProxyFinder proxy, final HostKeyCallback callback, final LoginCallback prompt, final CancelCallback cancel) throws BackgroundException {
         try {
             final CustomTrustSSLProtocolSocketFactory f
-                    = new CustomTrustSSLProtocolSocketFactory(trust, key, preferences.getProperty("connection.ssl.protocols.ftp").split(","));
+                    = new CustomTrustSSLProtocolSocketFactory(trust, key);
 
             final LoggingProtocolCommandListener listener = new LoggingProtocolCommandListener(this);
             final FTPClient client = new FTPClient(host.getProtocol(), f, f.getSSLContext()) {


ylangisc:
**Windows/IKVM**. IKVM 8.13 tracks jdk8u462, whose TLS 1.3 support is the JDK 11-era backport (JDK-8248721) — the implementation where the `sessionHostPortCache` reflection hack copies a session *without* the PSK ticket, so data channel resumption fails and servers requiring reuse (vsftpd `require_ssl_reuse`, ProFTPD, FileZilla Server — all defaults) reject the data connection after a TLS 1.3 handshake. The JDK 15+ ticket handling improvements were never backported to 8u. This should be verified on Windows against e.g. vsftpd with `require_ssl_reuse=YES` (login, list, up-/download). If it breaks, options are keeping `connection.ssl.protocols.ftp` with the full list as default and capping to TLS 1.2 only in the Windows defaults, or extending the workaround to fetch the PSK-bearing session copy from the cache under the control connection's host:port key.

- [ ] Needs to be tested on Windows and session re-use

ylangisc:
## TLS 1.3 breaks the FTPS data channel session reuse workaround

Re-enabling TLS 1.3 for FTPS (this PR, revert of c0524b78) is not sufficient on its own: with TLS 1.3 the data connection no longer resumes the TLS session of the control connection, so any server requiring session reuse on the data channel rejects transfers — `522 SSL connection failed; session reuse required` in vsftpd, `425 Unable to build data connection` in ProFTPD's `mod_tls`.

### Background

JSSE keys its client session cache by the peer `host:port` of the connection being opened and offers no API to bind an existing session to a new socket — there is no equivalent of OpenSSL's `SSL_set_session()` or GnuTLS' `gnutls_session_set_data()`. Since the FTPS data connection goes to a different port, the automatic lookup misses and JSSE performs a full handshake with a new session.

`FTPClient._prepareDataSocket_` has worked around this since 2013 (a9a6621789) by reflecting into `sun.security.ssl.SSLSessionContextImpl.sessionHostPortCache` and planting the control connection's session under the data connection's `host:port`, so the JDK's own resumption logic picks it up.

### What TLS 1.3 changes

The secret required for resumption no longer belongs to the handshake session. It arrives *after* the handshake in a `NewSessionTicket`, and the JDK attaches the resulting pre-shared key to a **copy** of the session stored in the client cache. `SSLSocket.getSession()` keeps returning the original object, which has no PSK.

The workaround therefore plants an unusable session:

```
javax.net.ssl|DEBUG|PreSharedKeyExtension.java:677|Existing session has no PSK.
javax.net.ssl|DEBUG|SSLExtensions.java:272|Ignore, context unavailable extension: pre_shared_key
```

The result is a full handshake on the data channel, and the failure is invisible on our side — nothing is logged, the user only sees the server's error reply.

### Measured on Temurin 25

In-process Apache MINA FTP server, explicit FTPS, one shared `SSLContext` for control and data channel:

| Setup | Data connection |
| --- | --- |
| TLS 1.2 | resumes the control session, as before |
| TLS 1.3, session from `getSession()` (current code) | full handshake |
| TLS 1.3, session from the cache | resumes (`Resuming session`, `Using PSK to derive early secret`), incl. three consecutive data connections |

### No supported alternative exists

Layering TLS onto the connected plain data socket and passing the control connection's host and port to `SSLSocketFactory.createSocket(Socket, host, port, autoClose)` does **not** move the cache key — the lookup uses the actual peer port — and yields `No session to resume` for TLS 1.2 as well. `commons-net` 3.13.0 exposes nothing for data channel reuse beyond the `_prepareDataSocket_` hook. The reflection workaround has to stay; it only needs to plant the right session object.

### Fix

In `_prepareDataSocket_`, look up the cache entry for the control connection's `host:port` and plant that session, falling back to the session of the control socket when the cache holds nothing (previous behaviour). `isValid()` now applies to the session actually being planted, so a control session reported as not rejoinable no longer prevents the cache lookup. Added a guard against a `null` `SSLSessionContext`, which could previously cause an NPE on an invalidated session.

Verified with a new parameterized test `FTPTLSSessionReuseTest` (TLSv1.2 and TLSv1.3, one and three consecutive data connections) asserting resumption by counting certificate validations in the client — a resumed handshake, both the abbreviated handshake in TLS 1.2 and PSK resumption in TLS 1.3, sends no certificate message.

- With the fix: `Tests run: 4, Failures: 0`
- Without the fix: `Tests run: 4, Failures: 2` (both TLS 1.3 cases)
- Full `ftp` module: `Tests run: 87, Failures: 0, Errors: 0, Skipped: 2`

### Still open

- [x] Verify against real servers instead of JSSE-to-JSSE. The number of possible resumptions depends on how many session tickets the server issues, since a PSK is single use. ProFTPD ships `TLSSessionTickets off` by default, which makes TLS 1.3 resumption impossible server-side. Needs a run against vsftpd (`require_ssl_reuse=YES`), ProFTPD and FileZilla Server, covering listing, upload, download and several consecutive transfers.
- [x] Windows build runs on IKVM, a different runtime with a different internal layout of the session cache. All measurements above are Temurin 25.
- [x] The workaround depends on `--add-opens java.base/sun.security.ssl` and `java.base/sun.security.util` plus JDK internals, and it degrades silently. Consider logging at warn level when reuse is required but cannot be set up.

ylangisc:
### Verified against real servers

Client run per protocol: three `LIST`, one upload, one download, one `LIST` — six data connections. Resumption is counted client-side as certificate validations (a resumed handshake sends no certificate) and confirmed server-side from the server's own TLS log.

#### vsftpd 3.0.3 (Debian bookworm, OpenSSL 3.0), `require_ssl_reuse=YES`

Server offers TLS 1.3 (`openssl s_client -starttls ftp -tls1_3` → `TLS_AES_256_GCM_SHA384`).

| Client | TLS 1.2 | TLS 1.3 |
| --- | --- | --- |
| with fix | all 6 data connections resumed | all 6 data connections resumed, 3/3 runs |
| without fix | all 6 data connections resumed | first 3 `LIST` resumed, upload does a full handshake, next data connection refused with `Connection closed without indication`, 3/3 runs |

The JSSE log of the failing run shows the transition: two `NewSessionTicket` messages at login (OpenSSL issues two), pre-shared keys get consumed, then `PreSharedKeyExtension: Existing session has no PSK` → full handshake → abort. Side note: vsftpd segfaults at that point (container exit 139), which reaches the user as a dropped connection.

#### ProFTPD 1.3.8 / mod_tls 2.9.2, session reuse required (default)

| Client | TLS 1.2 | TLS 1.3 |
| --- | --- | --- |
| with fix | resumed, `[stat]: SSL/TLS sessions resumed: 6` | resumed, `TLSv1.3 data connection accepted, using cipher TLS_AES_256_GCM_SHA384 (256 bits, resumed session)`, `[stat]: SSL/TLS sessions resumed: 6`, zero rejections |
| without fix | resumed | `client did not reuse TLS session, rejecting data connection (see the NoSessionReuseRequired TLSOptions parameter)`, and the client returns an **empty download** with no error — silent data loss |

Two concerns from the earlier analysis are settled by these runs:

- **`TLSSessionTickets off` is not a blocker.** With the directive set to `off` (verified inside the container), mod_tls still logs `SSLv3/TLS write session ticket` and resumption works unchanged — OpenSSL issues TLS 1.3 tickets regardless. Both protocols pass.
- **Multiple consecutive data connections are fine** on both servers: the cache entry is refreshed by the tickets the server sends after each connection.

Because the ProFTPD failure mode is silent, `_prepareDataSocket_`/`_openDataConnection_` now compare the session identifier of the data connection against the session planted in the cache and log a warning when they differ.

Not covered: FileZilla Server (GnuTLS instead of OpenSSL, different ticket behaviour) and the Windows build on IKVM. All measurements above are Temurin 25 against Linux servers.

ylangisc:
### Verified on the Windows build (IKVM, OpenJDK 8 class library)

Driven through `duck.exe` from `cli/windows/target/x64` — same IKVM runtime and the same `FTPClient` code as the shipped app. Preferences were set per run through environment variables, since `TerminalPreferences.getProperty` consults `System.getenv` first:

`connection.ssl.protocols=TLSv1.3 duck.exe --nokeychain --list ftps://…`

#### ProFTPD 1.3.8 / mod_tls 2.9.2, session reuse required (default)

| duck.exe | list | upload | longlist | download | server log |
| --- | --- | --- | --- | --- | --- |
| TLS 1.3 | ok | ok | ok | content correct | 4 × `TLSv1.3 data connection accepted, using cipher TLS_AES_256_GCM_SHA384 (256 bits, resumed session)`, 0 rejections |
| TLS 1.2 | ok | ok | ok | content correct | 4 × `TLSv1.2 data connection accepted, using cipher ECDHE-RSA-AES256-GCM-SHA384 (256 bits, resumed session)`, 0 rejections |

No client-side warnings in any run — in particular no `No field sessionHostPortCache in SSLSessionContext` and no `does not resume`. So the IKVM runtime negotiates TLS 1.3 (`TLSv1.3 connection accepted` on the control channel) and the reflection into the JSSE session cache works there as it does on the JVM.

**Control run** with `ftp.tls.session.requirereuse=false`: the server logs `client did not reuse TLS session, rejecting data connection (see the NoSessionReuseRequired TLSOptions parameter)`, confirming that planting the session in the cache is what makes the Windows client work, and that the rejection is silent for the user — which is what the added warning is for.

#### vsftpd 3.0.3, `require_ssl_reuse=YES`

TLS 1.3 listing and download verified from the Windows build, downloaded content correct, no client warnings. Uploads only succeed with `ftp.datachannel.epsv=true` in this container setup, and vsftpd 3.0.3 segfaults after TLS 1.3 sessions — a server-side crash also seen with the JVM client, unrelated to the client code. TLS 1.2 against this particular server could not be exercised from the Windows build for a reason unrelated to this change.

#### Not covered

The pre-fix behaviour on Windows was not measured; that would need an IKVM rebuild of 8faf4d9e5d. All Windows runs above are with the fix, against Linux servers in Docker.

ylangisc (approved):

PR #18421, opened 2026-09-08
 Consume response body on error

### LINKED ISSUE #18398: Review connection handling for error responses
Several features executing requests with the raw Apache HttpClient check the response status and throw a mapped exception on error without consuming the response entity. The `HttpMethodReleaseInputStream` wrapping (which releases the connection on close) is only applied on the success path. On the error path the entity is never read, so the underlying connection is not returned to the connection pool until garbage collection. Long running sessions issuing many failing requests can exhaust the pool.

Found while addressing the same defect in the review of #18397.

### Affected

All six read features share the identical block:

```java
final HttpResponse response = client.execute(request);
switch(response.getStatusLine().getStatusCode()) {
    case HttpStatus.SC_OK:
    case HttpStatus.SC_PARTIAL_CONTENT:
        return new HttpMethodReleaseInputStream(response, status);
    default:
        // response entity never consumed
        throw new DefaultHttpResponseExceptionMappingService().map(...);
```

- `googledrive` — `DriveReadFeature#getStream` (DriveReadFeature.java:153)
- `brick` — `BrickReadFeature#read` (BrickReadFeature.java:81)
- `dracoon` — `SDSReadFeature#read` (SDSReadFeature.java:85)
- `deepbox` — `DeepboxReadFeature#read` (DeepboxReadFeature.java:192)
- `storegate` — `StoregateReadFeature#read` (StoregateReadFeature.java:80)
- `eue` — `EueReadFeature#read` (EueReadFeature.java:83)

Related variants in the `eue` module where the response is not released at all:

- `EueSession#login` service discovery (EueSession.java:169–183): error path not consumed; the success path reads the entity through `JsonParser.parseReader` without closing the stream or consuming the entity either
- `EueSession#login` PACS request (EueSession.java:188): return value of `execute` discarded, response never consumed
- `EueWriteFeature#cancel` (EueWriteFeature.java:161): response of the `HttpDelete` discarded, never consumed

### Proposed fix

Consume the entity before throwing:

```java
default:
    EntityUtils.consumeQuietly(response.getEntity());
    throw new DefaultHttpResponseExceptionMappingService().map(...);
}
```

`consumeQuietly` rather than `consume` so an `IOException` while draining the error body cannot displace the status-based exception with a generic mapping from the surrounding `catch(IOException)`. For the discarded responses in the `eue` module, consume (or close) the response after `execute`.

### Verified not affected

- Write features consuming in `finally`: `DriveWriteFeature`, `GoogleStorageWriteFeature`, `BrickWriteFeature`, `EueWriteFeature` upload path, `EueMultipartWriteFeature`
- `ResponseHandler`-style `execute(request, handler)` calls which release the connection internally: `DeepboxWriteFeature`, `NextcloudSession`, `OwncloudSession`, `CteraAuthenticationHandler`
- `DAVReadFeature`: `SardineException` is an `IOException`; the catch calls `request.abort()`, discarding the connection instead of reusing it — wasteful but no leak
- Library-managed error handling: S3 (jets3t), OpenStack Swift (joss), Backblaze B2 SDK, Dropbox SDK executor (`DropboxCommonsHttpRequestExecutor`), OneDrive client executor (`GraphCommonsHttpRequestExecutor`)
- `BoxReadFeature`, `CteraDirectIOReadFeature`: no status check at all — the entity is always wrapped, so no connection leak. An error body would however be read as file content; that is a correctness problem out of scope here, to be tracked in a separate issue

dkocher:
Fix #18398.

### DIFF
--- a/brick/src/main/java/ch/cyberduck/core/brick/BrickReadFeature.java
+++ b/brick/src/main/java/ch/cyberduck/core/brick/BrickReadFeature.java
@@ -36,6 +36,7 @@ import org.apache.http.client.HttpResponseException;
 import org.apache.http.client.methods.HttpGet;
 import org.apache.http.client.methods.HttpUriRequest;
 import org.apache.http.message.BasicHeader;
+import org.apache.http.util.EntityUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
@@ -79,6 +80,7 @@ public class BrickReadFeature implements Read {
                 case HttpStatus.SC_PARTIAL_CONTENT:
                     return new HttpMethodReleaseInputStream(response, status);
                 default:
+                    EntityUtils.consumeQuietly(response.getEntity());
                     throw new DefaultHttpResponseExceptionMappingService().map("Download {0} failed", new HttpResponseException(
                         response.getStatusLine().getStatusCode(), response.getStatusLine().getReasonPhrase()), file);
             }
--- a/storegate/src/main/java/ch/cyberduck/core/storegate/StoregateMultipartWriteFeature.java
+++ b/storegate/src/main/java/ch/cyberduck/core/storegate/StoregateMultipartWriteFeature.java
@@ -158,7 +158,7 @@ public class StoregateMultipartWriteFeature implements MultipartWrite<File> {
                                 throw e;
                             }
                             finally {
-                                EntityUtils.consume(response.getEntity());
+                                EntityUtils.consumeQuietly(response.getEntity());
                             }
                         }
                         catch(IOException e) {
@@ -221,7 +221,7 @@ public class StoregateMultipartWriteFeature implements MultipartWrite<File> {
                         throw new IOException(e);
                     }
                     finally {
-                        EntityUtils.consume(response.getEntity());
+                        EntityUtils.consumeQuietly(response.getEntity());
                     }
                 }
             }
--- a/storegate/src/main/java/ch/cyberduck/core/storegate/StoregateMultipartWriteFeature.java
+++ b/storegate/src/main/java/ch/cyberduck/core/storegate/StoregateMultipartWriteFeature.java
@@ -39,6 +39,7 @@ import org.apache.http.HttpResponse;
 import org.apache.http.HttpStatus;
 import org.apache.http.client.entity.EntityBuilder;
 import org.apache.http.client.methods.HttpPut;
+import org.apache.http.entity.BufferedHttpEntity;
 import org.apache.http.util.EntityUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -146,6 +147,7 @@ public class StoregateMultipartWriteFeature implements MultipartWrite<File> {
                                         offset += content.length;
                                         break;
                                     default:
+                                        EntityUtils.updateEntity(response, new BufferedHttpEntity(entity));
                                         final ApiException failure = new ApiException(response.getStatusLine().getStatusCode(),
                                                 response.getStatusLine().getReasonPhrase(), Collections.emptyMap(),
                                                 EntityUtils.toString(response.getEntity()));
@@ -211,6 +213,7 @@ public class StoregateMultipartWriteFeature implements MultipartWrite<File> {
                             case HttpStatus.SC_NO_CONTENT:
                                 break;
                             default:
+                                EntityUtils.updateEntity(response, new BufferedHttpEntity(response.getEntity()));
                                 final ApiException failure = new ApiException(response.getStatusLine().getStatusCode(),
                                         response.getStatusLine().getReasonPhrase(), Collections.emptyMap(),
                                         EntityUtils.toString(response.getEntity()));
--- a/storegate/src/main/java/ch/cyberduck/core/storegate/StoregateMultipartWriteFeature.java
+++ b/storegate/src/main/java/ch/cyberduck/core/storegate/StoregateMultipartWriteFeature.java
@@ -147,7 +147,6 @@ public class StoregateMultipartWriteFeature implements MultipartWrite<File> {
                                         offset += content.length;
                                         break;
                                     default:
-                                        EntityUtils.updateEntity(response, new BufferedHttpEntity(entity));
                                         final ApiException failure = new ApiException(response.getStatusLine().getStatusCode(),
                                                 response.getStatusLine().getReasonPhrase(), Collections.emptyMap(),
                                                 EntityUtils.toString(response.getEntity()));
--- a/storegate/src/main/java/ch/cyberduck/core/storegate/StoregateMultipartWriteFeature.java
+++ b/storegate/src/main/java/ch/cyberduck/core/storegate/StoregateMultipartWriteFeature.java
@@ -39,7 +39,6 @@ import org.apache.http.HttpResponse;
 import org.apache.http.HttpStatus;
 import org.apache.http.client.entity.EntityBuilder;
 import org.apache.http.client.methods.HttpPut;
-import org.apache.http.entity.BufferedHttpEntity;
 import org.apache.http.util.EntityUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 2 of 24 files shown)


storegate/src/main/java/ch/cyberduck/core/storegate/StoregateSession.java
```diff
@@ -148,17 +148,19 @@ public void login(final LoginCallback controller, final CancelCallback cancel) t
                         }
                         break;
                     case HttpStatus.SC_FORBIDDEN:
+                        EntityUtils.updateEntity(response, response.getEntity());
```
ylangisc, line 151: `EntityUtils.updateEntity(response, entity)` is implemented as `consume(response.getEntity()); response.setEntity(entity);`. Passing the response's own entity back in drains and closes the stream and then re-installs the same, now-closed entity. `BasicHttpEntity#getContent` returns the same stream object, so the `EntityUtils.toString(response.getEntity())` two lines below throws `IOException("Attempted read from closed stream.")`.

That `IOException` lands in the `catch(IOException e)` at the bottom of `login()` and is mapped by `DefaultIOExceptionMappingService`, so a 403 no longer produces the `LoginFailureException` with the server's message (and the re-login prompt), but a generic I/O error. Same for the `default:` branch at line 157. On master `toString` ran before the `finally { consume }`, so this is a regression.

Suggest simply removing both `updateEntity` lines: `toString` already reads to EOF and closes the stream, and the `consumeQuietly` in `finally` covers the rest.

ylangisc (commented):
Two more of the same class that are outside the diff hunks, so noting them here instead of inline:

- `StoregateWriteFeature#cancel` (line 204): `session.getClient().getClient().execute(delete)` discards the response, so the entity is never consumed and the connection stays checked out until GC. This is the same defect the PR fixes in `EueWriteFeature#cancel`, and `cancel()` runs on every failed upload path (including inside the `DefaultRetryCallable` in `StoregateMultipartWriteFeature`), so it leaks one connection per retry. Same one-liner fix: `EntityUtils.consumeQuietly(session.getClient().getClient().execute(delete).getEntity())`.
- `SpectraBulkService#clear` (line 338-341): raw `execute` on `_rest_/cache_filesystem?reclaim` throws `HttpResponseException` on non-204 without consuming the entity. Low frequency (once per bulk operation), but neither fixed here nor listed under "verified not affected" in #18398.

storegate/src/main/java/ch/cyberduck/core/storegate/StoregateWriteFeature.java
```diff
@@ -116,6 +116,7 @@ public File call(final HttpEntity entity) throws BackgroundException {
                                 fileid.cache(file, result.getId());
                                 return result;
                             default:
+                                EntityUtils.updateEntity(putResponse, putResponse.getEntity());
```
ylangisc: Same issue as in `StoregateSession#login`: `updateEntity(putResponse, putResponse.getEntity())` consumes the error body, so `EntityUtils.toString(putResponse.getEntity())` below throws `IOException`. That is not a `BackgroundException`, so it falls through to the `catch(IOException e)` and is mapped via `HttpExceptionMappingService` as a generic I/O failure. The status-based mapping in `StoregateExceptionMappingService` (404 → `fileid.cache(file, null)`, conflict/quota mapping, server `message`) never runs. Dropping the `updateEntity` line restores the previous behaviour.

storegate/src/main/java/ch/cyberduck/core/storegate/StoregateWriteFeature.java
```diff
@@ -174,25 +175,22 @@ protected String start(final Path file, final TransferStatus status) throws Back
...
+                    if(response.containsHeader(HttpHeaders.LOCATION)) {
+                        EntityUtils.consumeQuietly(response.getEntity());
+                        return response.getFirstHeader(HttpHeaders.LOCATION).getValue();
+                    }
+                    EntityUtils.updateEntity(response, response.getEntity());
+                    throw new StoregateExceptionMappingService(fileid).map("Upload {0} failed",
+                            new ApiException(response.getStatusLine().getStatusCode(), response.getStatusLine().getReasonPhrase(), Collections.emptyMap(),
+                                    EntityUtils.toString(response.getEntity())), file);
+                default:
+                    EntityUtils.updateEntity(response, response.getEntity());
```
ylangisc: Same `updateEntity(response, response.getEntity())` problem on both error branches here (184 and 189): `toString` throws on the closed stream and the exception becomes a generic `HttpExceptionMappingService` mapping from the `catch(IOException)`.

This one has a visible functional consequence: `write()` relies on `start()` throwing `InteroperabilityException` to retry without the lock id (`catch(InteroperabilityException e) { ... start(file, status.setLockId(null)); }`). With the `IOException` mapping that fallback never triggers, so uploads that used to recover now fail outright with a misleading connection error.

Also, the old `try { switch } finally { consume }` was dropped, so connection release on the error path now depends solely on `toString` succeeding. Suggest keeping the `try/finally { consumeQuietly }` around the switch, letting `SC_OK`-without-`Location` fall through to `default:` (the two throw blocks are identical), and removing the `updateEntity` calls.

storegate/src/main/java/ch/cyberduck/core/storegate/StoregateMultipartWriteFeature.java
```diff
@@ -211,6 +213,7 @@ public void close() throws IOException {
                             case HttpStatus.SC_NO_CONTENT:
                                 break;
                             default:
+                                EntityUtils.updateEntity(response, new BufferedHttpEntity(response.getEntity()));
```
ylangisc: Nit: this variant is correct (it buffers the response entity), but the buffering is redundant. The body is read exactly once by `EntityUtils.toString`, which reads to EOF and closes the stream itself, and `consumeQuietly` runs in `finally`. `BufferedHttpEntity` + `updateEntity` just add a second in-memory copy of the error body and an extra import. `EntityUtils.toString(response.getEntity())` alone does the job, and using that form everywhere would also have avoided the broken copies at line 150 and in `StoregateSession`/`StoregateWriteFeature`.

brick/src/main/java/ch/cyberduck/core/brick/BrickReadFeature.java
```diff
@@ -79,6 +80,7 @@ public InputStream read(final Path file, final TransferStatus status, final Conn
                 case HttpStatus.SC_PARTIAL_CONTENT:
                     return new HttpMethodReleaseInputStream(response, status);
                 default:
+                    EntityUtils.consumeQuietly(response.getEntity());
```
ylangisc, line 83: Not blocking, but this identical `consumeQuietly` + `throw new DefaultHttpResponseExceptionMappingService().map("Download {0} failed", new HttpResponseException(...), file)` block is now pasted into six read features (Brick, Deepbox, SDS, EUE, Drive, Storegate) plus `EueSession`. The next `<Protocol>ReadFeature` will copy the switch without the consume line and reintroduce #18398.

An overload on `HttpResponseExceptionMappingService`, e.g. `map(String message, HttpResponse response, Path file)` that does the `consumeQuietly` and builds the `HttpResponseException` from the status line, would collapse each site to a single `throw ... map("Download {0} failed", response, file)` and put the release where the status line is turned into an exception.

storegate/src/main/java/ch/cyberduck/core/storegate/StoregateSession.java
ylangisc: In 4d8c8b48c746fcbdaeaa812d62c423077b682c9c.

storegate/src/main/java/ch/cyberduck/core/storegate/StoregateWriteFeature.java
ylangisc: In 4d8c8b48c746fcbdaeaa812d62c423077b682c9c.

storegate/src/main/java/ch/cyberduck/core/storegate/StoregateMultipartWriteFeature.java
ylangisc: In 4d8c8b48c746fcbdaeaa812d62c423077b682c9c.

storegate/src/main/java/ch/cyberduck/core/storegate/StoregateWriteFeature.java
ylangisc: In 4d8c8b48c746fcbdaeaa812d62c423077b682c9c.

storegate/src/main/java/ch/cyberduck/core/storegate/StoregateMultipartWriteFeature.java
```diff
@@ -146,6 +147,7 @@ public Void call() throws BackgroundException {
                                         offset += content.length;
                                         break;
                                     default:
+                                        EntityUtils.updateEntity(response, new BufferedHttpEntity(entity));
```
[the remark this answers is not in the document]
ylangisc: In 24e1c1f61ded5f6567375745178ad9e9b070ebb2.

ylangisc (approved):

Issue #14166, opened 2023-01-31
Add support for oCIS - ownCloud Infinite Scale


dkocher:
New concept of spaces accessible using [LibreGraph](https://github.com/owncloud/libre-graph-api)

- [ ] Generate from [Swagger Spec](https://owncloud.dev/libre-graph-api/) the implementation to retrieve drives for a user
- [ ] Reuse our current Microsoft Graph Client [implementation](https://github.com/iterate-ch/onedrive-java-client).

oCIS uses WebFinger [RFC7033](http://tools.ietf.org/html/rfc7033) to locate the oCIS instance for an End-User.

- [ ] Discovery of endpoint with https://owncloud.dev/ocis/storage_registry_discovery/

Now uses OAuth for authentication. Must discover the OAuth endpoints somehow.

- [ ] Obtain OpenID configuration from `/.well-known/openid-configuration`?

dkocher:
With #14877 we allow to configure usage of OAuth for WebDAV connections in connection profiles.

michaelstingl:
### Legacy OAuth 2.0 vs Basic Auth detection (still works in oC10 + oCIS as of today)

In an unauthenticated PROPFIND to the WebDAV endpoint, ownCloud clients check for `Www-Authenticate: Bearer` header.

```bash
curl -s -I --http1.1 "https://demo.owncloud.org/remote.php/dav/files/" | grep -i -e "HTTP/" -e "www-authenticate"
HTTP/1.1 401 Unauthorized
Www-Authenticate: Basic realm="ownCloud", charset="UTF-8"
```

➡️ ownCloud clients use Basic Auth

```bash
curl -s -I --http1.1 "https://demo.owncloud.com/remote.php/dav/files/" | grep -i -e "HTTP/" -e "www-authenticate"
HTTP/1.1 401 Unauthorized
Www-Authenticate: Bearer realm="ownCloud"
Www-Authenticate: Basic realm="ownCloud", charset="UTF-8"
```

➡️ ownCloud clients use OAuth 2.0 (if there's no trace of OpenID connect)
➡️ open `/index.php/apps/oauth2/authorize` in web UI

Here you can find more information:
- https://github.com/owncloud/oauth2/wiki/OAuth-code-Flow-Sequence-Diagram

### Legacy OpenID Connect detection

After `Www-Authenticate: Bearer` was detected, ownCloud clients check for `/.well-known/openid-configuration` on that instance, mirrored from the IdP.

[...]

PR #18448, opened 2026-09-15
Assign port to loopback redirect URI before authorization request.

### LINKED ISSUE #18447: OAuth login fails with loopback redirect URI without port
**Describe the bug**
A connection profile with `OAuth Redirect Url` set to a loopback address without a port, e.g. `http://localhost/`, cannot complete the OAuth login.

- `OAuth2AuthorizationService` builds the authorization URL with the redirect URI from the profile unchanged and uses the same unchanged value in the token request.
- `LoopbackOAuth2AuthorizationCodeProvider` binds the callback `HttpServer` to a random port when the redirect URI has no port (9023b9b8bb "Assign random port number when not explicitly set in redirect URI"), but the bound port is never propagated to the `redirect_uri` parameter of the already built authorization URL or to the token exchange.

The identity provider therefore redirects the browser to `http://localhost/?code=…` (port 80) where nothing is listening.

**To Reproduce**
1. Use a connection profile with an OAuth authorization code flow and `OAuth Redirect Url` set to `http://localhost/`.
2. Connect and log in with the identity provider in the web browser.
3. The browser is redirected to `http://localhost/?code=…`, which fails to load, and the login in Cyberduck never completes.

**Expected behavior**
Per [RFC 8252 section 7.3](https://datatracker.ietf.org/doc/html/rfc8252#section-7.3), the client picks the port at request time. Microsoft Entra ID ignores the port for `localhost`, and Keycloak matches loopback redirects regardless of port, so a random port works as long as the same effective redirect URI is sent in both the authorization request and the token request.

Redirect URIs with an explicit port, e.g. `http://localhost:8765/`, work.

**Affected versions**
`master` and 9.5.0 through 9.5.4.

**Additional context**
Proposed fix: when the redirect URI is a loopback address without a port, allocate a free port before building the authorization URL, similar to `RegisterClientOAuth2RequestInterceptor` (AWS IAM Identity Center), and use that effective redirect URI for the authorization URL, the loopback callback server and the token request.

dkocher:
Fix #18447.

When the OAuth redirect URI is a loopback address without a port (e.g. `http://localhost/`), allocate a free port before building the authorization URL and use the same effective redirect URI for the authorization request, the loopback callback server and the token request (RFC 8252 section 7.3).


dkocher:
Relates to #14166.

ylangisc (approved):

PR #18458, opened 2026-09-16
Request server-side checksum on upload with irods protocol and verify against local file

### LINKED ISSUE #18457: iRODS uploads never register a checksum in the catalog
iRODS uploads never register a checksum in the catalog, even with "Verify checksum" enabled in Preferences, Transfers, Checksum. The transfer log shows "Calculate checksum for …" but the data object ends up without a checksum.

 - To reproduce: connect to an iRODS 5 server with the iRODS profile, enable the checksum preference, upload a file, run ils -L on the
    object. Compare with iput -K, which does register one.
  - Expected behavior: the server registers a checksum, and Cyberduck verifies it against the local file, as it does for S3.
  - Desktop: Fedora core 43 and Cyberduck 9.5.4 (45528).
  - Additional context: the root cause. The upload feature closes the replica without the compute_checksum instruction in the replica_close
    call, and the iRODS write feature does not override Write.checksum(), so the preference has no effect.

A patch follows as a pull request.
I attach the logs of cyberduck, before and after the patch and the output of irods file listing.
  - Stock log, line 1609: "Calculate checksum for test.bin…" appears, then nothing checksum-related follows before "Upload complete" at
    line 3332.
  - Patched log, line 1722: "compute checksum = [true]", and line 3488: the server-registered SHA-256 matches the local file.

[cyberduck-9.5.4-patched-upload.log](https://github.com/user-attachments/files/32300462/cyberduck-9.5.4-patched-upload.log)
[cyberduck-9.5.4-stock-upload.log](https://github.com/user-attachments/files/32300461/cyberduck-9.5.4-stock-upload.log)
[ils-L-after-patched.txt](https://github.com/user-attachments/files/32300464/ils-L-after-patched.txt)
[ils-L-after-stock-9.5.4.txt](https://github.com/user-attachments/files/32300463/ils-L-after-stock-9.5.4.txt)

ccacciari:
Fix #18457

Request compute_checksum on close of the primary stream when a local checksum was computed, and return a SHA-256 checksum from the write feature so verification runs.


irods/src/main/java/ch/cyberduck/core/irods/IRODSUploadFeature.java
```diff
@@ -123,21 +136,27 @@ public Void upload(final Write<Void> write, final Path file, final Local local,
...
+                        // Only request a checksum if the transfer completed.
+                        out.close(closeInstructions(setCompletionFlag && computeChecksum));
 
                         if(setCompletionFlag) {
                             status.setComplete();
                         }
                     }
                 }
+
+                if(setCompletionFlag) {
```
dkocher: Condition 'setCompletionFlag' is always 'true'

irods/src/main/java/ch/cyberduck/core/irods/IRODSUploadFeature.java
```diff
@@ -269,7 +292,46 @@ public Write.Append append(Path file, TransferStatus status) throws BackgroundEx
...
+     * Compare the checksum registered by the server with the checksum of the local file
+     * computed prior to the transfer, if any.
+     */
+    private void verifyChecksum(final Path file, final TransferStatus status) throws BackgroundException {
+        if(!hasChecksum(status)) {
+            log.debug("no local checksum available for [{}]. skipping verification.", file);
+            return;
+        }
+        final Checksum expected = status.getChecksum();
+        final PathAttributes attributes = new IRODSAttributesFinderFeature(session).find(file, new DisabledListProgressListener());
```
dkocher: It might be expensive to query the file attributes for every file uploaded. If that is a non-issue, then make the `Upload` and `Write` features return the latest attributes after upload from the server as `List<String>`. In conjunction with `AttributesAdapter<List<String>>` implemented already, that will allow to save another round-trip to the server.

dkocher (changes requested):

irods/src/main/java/ch/cyberduck/core/irods/IRODSUploadFeature.java
ccacciari: Right, after the try-with-resources the flag can only be true (cancel returns early, errors propagate). Removed in ef892f2.

irods/src/main/java/ch/cyberduck/core/irods/IRODSUploadFeature.java
ccacciari: Agreed, find() was two round-trips (stat + GenQuery2) and the stat is redundant right after a write. In ef892f2 IRODSUploadFeature/IRODSWriteFeature are now Upload<List<String>>/Write<List<String>>: after the primary replica is closed, a single GenQuery2 (shared with IRODSAttributesFinderFeature) fetches the row, which is set via status.setResponse(...) using the existing toAttributes and used to verify the local checksum.
Since replica_close returns nothing in irods4j, one query is the minimum to obtain the server-registered checksum. I made it unconditional so the client always gets current size/mtime/checksum after an upload like the other protocols; happy to restrict it to when checksum verification is enabled if you prefer. Verified against iRODS 5.0.2 through both the single-stream and the parallel code paths.

dkocher (approved):

